PR #28, opened 2026-01-11
docs: suggest qwen3-embedding and embeddinggemma

jedisct1:
## Description

Include successors to nomic and mxbai.

`qwen3-embedding` is currently the best performing embedding model.

It's also pretty good at code comprehension. Not the fastest option, but great to do one-shot analysis of code bases.

`embeddinggemma` only requires about 200 MB memory, but is fast, and performs really well.

Both have performance comparable to 8B models.

## Type of Change

- [ ] New feature (non-breaking change that adds functionality)
- [X] Documentation update
- [ ] Performance improvement
- [ ] Test update

### DIFF
--- a/docs/src/content/docs/backends/embedders.md
+++ b/docs/src/content/docs/backends/embedders.md
@@ -53,6 +53,8 @@ embedder:
 | `nomic-embed-text` | 768 | Fast | Good |
 | `mxbai-embed-large` | 1024 | Medium | Better |
 | `all-minilm` | 384 | Very Fast | Basic |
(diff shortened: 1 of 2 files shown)


yoanbernabeu (approved):
Thanks Frank! 🙏

Great additions to the documentation - `qwen3-embedding` and `embeddinggemma` are indeed excellent modern alternatives to the existing models. The performance notes you've included (especially the 200MB footprint for embeddinggemma) are super helpful for users making their choice.

Really appreciate you taking the time to contribute and keep the docs up-to-date with the latest embedding models!

PR #100, opened 2026-02-02
feat(cli): add non-interactive workspace create and mcp-serve --workspace flag

jugrajsingh:
## Description

Add CLI flags (`--name`, `--backend`, `--embedder-provider`, `--embedder-model`, `--dsn`) to `workspace create` for scripted/CI usage, enabling fully non-interactive workspace creation without TUI prompts. Add `--workspace` flag to `mcp-serve` so MCP tools can be scoped to a specific workspace. Add `FindWorkspaceConfig()` and `WorkspaceStoreConfig()` helpers in the config package for programmatic workspace resolution.

Documentation updated for workspace mode, parallelism tiers, MCP workspace-scoped configuration, and watch daemon examples.

## Type of Change

- [ ] Bug fix (non-breaking change that fixes an issue)
- [x] New feature (non-breaking change that adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [x] Documentation update
- [ ] Refactoring (no functional changes)
- [ ] Performance improvement
- [x] Test update

## How Has This Been Tested?

- [x] Unit tests
- [ ] Integration tests
- [x] Manual testing

**Test Configuration:**
- OS: Fedora Linux (6.18.7-200.fc43.x86_64)
- Go version: 1.25.6
- Embedding provider: Ollama

## Checklist

- [x] My code follows the project's code style
- [x] I have run `golangci-lint run` and fixed any issues
- [x] I have added tests that prove my fix/feature works
- [x] I have updated the documentation if needed
- [x] I have added an entry to CHANGELOG.md (if applicable)
- [x] My changes generate no new warnings
- [x] All new and existing tests pass

## Screenshots (if applicable)

N/A

## Additional Notes

- All existing single-project workflows remain unchanged (100% backward compatible)
- New test files: `cli/mcp_serve_test.go`, `cli/workspace_create_test.go`, `config/workspace_test.go`

PR #104, opened 2026-02-04
fix(mcp): add titleFixWriter for Windsurf stdio compatibility

cmdaltctr:
This PR adds a `titleFixWriter` that wraps stdout in the MCP server to fix three issues with Windsurf (and potentially other IDE) integration:

1. **Missing newline after JSON-RPC responses** — The default `server.ServeStdio` writes JSON without a trailing newline. Clients relying on line-delimited JSON-RPC (like Windsurf) hang waiting for the message to complete.
2. **Tool title in wrong location** — Windsurf expects `title` at the tool root level, but `mcp-go` places it inside `annotations`. The writer moves it to the correct location.
3. **Missing `$schema` in `inputSchema`** — Windsurf requires `$schema` for tool discovery. The writer injects it if absent.

### Changes from previous version

This PR was previously scoped to also include a global-mode fallback for `mcp-serve`. That has been dropped since the upstream `--workspace` flag (#100) and `resolveMCPTarget` provide a better solution. This PR now focuses solely on the `titleFixWriter` fix in `mcp/server.go`.

### DIFF
--- a/mcp/server.go
+++ b/mcp/server.go
@@ -6,6 +6,8 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
+	"io"
+	"os"
 	"strings"
 
 	"github.com/alpkeskin/gotoon"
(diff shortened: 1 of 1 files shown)


yoanbernabeu:
Hi @cmdaltctr, thanks for this contribution!

The PR currently has merge conflicts with `main`. Could you rebase your branch on the latest `main` to resolve them?

Everything else looks good — CI is green (only `codecov/patch` fails due to coverage on new code, which is not blocking). We'll be happy to merge once the conflicts are resolved!

cmdaltctr:
Hi @yoanbernabeu, thanks for the review!

I've rebased onto the latest `main` and **scoped down the PR** to focus only on the `titleFixWriter` fix in `mcp/server.go`.

The global-mode fallback logic I originally included in `cli/mcp_serve.go` has been dropped — the `--workspace` flag and `resolveMCPTarget` from #100 provide a much better solution for that use case.

What remains is a single focused change: replacing `server.ServeStdio` with a custom writer that fixes three Windsurf compatibility issues (missing newline, title placement, $schema injection).

PR #124, opened 2026-02-08
Fix: grepai update fails on cross-device link and newline output

hansipie:
## Description

This commit addresses two issues:
1.  The 'invalid cross-device link' error when running 'grepai update' on filesystems where /tmp and the install directory are different. Implemented a safeRename function with copy-then-delete fallback.
2.  A minor UI issue where the error message would appear on the same line as the download progress. Ensured a newline is printed before error messages.

## Type of Change

- [x] Bug fix (non-breaking change that fixes an issue)
- [ ] New feature (non-breaking change that adds functionality)
- [ ] Documentation update
- [ ] Performance improvement
- [ ] Test update

## How Has This Been Tested?

- [ ] Unit tests
- [ ] Integration tests
- [x] Manual testing

**Test Configuration:**
- OS: Ubuntu
- Go version: 1.24.2

## Checklist

- [x] I have run `go vet`, `gofmt`, `go mod tidy && go mod verify`  and fixed any issues
- [ ] I have added tests that prove my fix/feature works
- [ ] I have updated the documentation if needed
- [ ] I have added an entry to CHANGELOG.md (if applicable)
- [ ] My changes generate no new warnings
- [ ] All new and existing tests pass

### DIFF
--- a/updater/cross_device_unix.go
+++ b/updater/cross_device_unix.go
@@ -1,0 +1,12 @@
+//go:build !windows
+
+package updater
+
+import (
+	"errors"
+	"syscall"
+)
+
+func isCrossDeviceError(err error) bool {
+	return errors.Is(err, syscall.EXDEV)
+}
--- a/updater/cross_device_windows.go
+++ b/updater/cross_device_windows.go
@@ -1,0 +1,7 @@
+//go:build windows
+
+package updater
+
+func isCrossDeviceError(err error) bool {
+	return false
+}
--- a/updater/updater.go
+++ b/updater/updater.go
@@ -469,12 +469,12 @@ func safeRename(oldpath, newpath string) error {
 		return nil
 	}
 
-	// Check if the error is "invalid cross-device link" (EXDEV)
-	// On Linux, this is often represented as "cross-device link"
-	// On other Unix-like systems, it might be slightly different, but the intent is the same.
-	if linkErr, ok := err.(*os.LinkError); ok && strings.Contains(linkErr.Error(), "cross-device link") {
-		// Fallback to copy and delete
-		// fmt.Printf("Warning: Rename failed with cross-device link error. Falling back to copy and delete for %s to %s\n", oldpath, newpath) // For debugging
+	// Fallback to copy+delete for cross-device link errors (EXDEV)
+	if isCrossDeviceError(err) {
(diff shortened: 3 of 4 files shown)


yoanbernabeu:
Thanks for this PR @hansipie! The cross-device link fix addresses a real issue for Linux users, and the progress line fix is a nice touch. A few suggestions before merging:

**1. Error detection relies on string matching (should fix)**

The current check `strings.Contains(linkErr.Error(), "cross-device link")` is fragile — it depends on the error message text which could vary across OS or locales. The idiomatic Go approach would be to use `errors.Is(err, syscall.EXDEV)` instead, which is more robust and portable.

**2. Hardcoded file permissions (nice to have)**

The destination file is created with fixed `0755` permissions. Ideally, you'd preserve the source file's permissions via `os.Stat()` before copying. Minor for a binary, but cleaner.

**3. No unit tests for `safeRename` (nice to have)**

This function is critical (it replaces the running binary) and is easily testable in isolation. A simple test covering at least the happy path would add confidence.

Point 1 is the main blocker — points 2 and 3 would be nice improvements but are not blocking. Great contribution overall!

hansipie:
## Changes Made

1. **updater/cross_device_unix.go (new)** — Added EXDEV detection via `errors.Is(err, syscall.EXDEV)` for all non-Windows platforms.
2. **updater/cross_device_windows.go (new)** — Added a stub returning `false` for Windows, where `syscall.EXDEV` is not defined.
3. **updater/updater.go line 475** — Replaced the fragile `strings.Contains(linkErr.Error(), "cross-device link")` with `isCrossDeviceError(err)`. Added `os.Stat` to preserve the source file permissions (`srcInfo.Mode()`) instead of using the hardcoded `0755`.
4. **updater/updater_test.go** — Added two tests:
* `TestSafeRename_HappyPath`: Verifies that the source is removed and the destination contains the correct data.
* `TestSafeRename_PermissionsPreserved`: Verifies that `0744` permissions are correctly preserved (skipped on Windows).

yoanbernabeu:
Thanks for the updates @hansipie! All three points from the first review have been properly addressed — the EXDEV detection via build tags is clean, permissions are now preserved, and the tests cover the essentials. LGTM, merging once CI is green 🚀

PR #131, opened 2026-02-09
First pass at adding --path filter

### LINKED ISSUE #130: [Bug]: Workspaces with projects are Completely Broken - doesn't pass project filter to SQL
### Bug Description

I have spent HOURS trying to get grepai to work and I'm just about to the point of giving up...

I am trying to set up workspaces and projects to workaround the fact that it's not possible to make grepai index folders listed in .gitignore.

After spending hours on this, anytime I do a search using --project, it runs for about 20 seconds and returns 

`No results found.`

When I remove the --project option, it does return results... But many have the wrong path. I see tons of "results" where the path is "project_root/project_root" 

Also: I checked the database queries to psql, and this is the query it's sending:

```
SELECT id, file_path, start_line, end_line, content, vector, hash, updated_at,
				1 - (vector <=> $1) as score
			FROM chunks
			WHERE project_id = $2
			ORDER BY vector <=> $1
			LIMIT $3
```

Where:
$1 is the huge vector list (redacted)
$2 = 'workspace:tymemud', 
$3 = '10' (I was sending "--limit 5")

But ALL of the entries in the "chunks" table have the same project_id:

```
grepai=> select distinct project_id from chunks;
    project_id     
-------------------
 workspace:tymemud
(1 row)
```

And when I re-run the exact same query that grepai is sending, I get results... but none of them are for the "file_path" that matches the project I'm filtering on.

It *looks* to me like what grepai is doing here is:

Searching the entire workspace for the "search" query. Asking for 2x whatever the user's filter count is (IE: I passed filter=5, it passed limit=10 to SQL) and then it does a post query filter looking for a file_path that matches the project.

However: This will not work in large codebases. I have 40 different project folders. So doing a "limit 10" means there's VERY little chance that one of those 10 rows returned from SQL will be for the specific project I'm looking for.

Why is it not filtering on "path" at the database level?

When I "fix" the query myself and so this:

```
SELECT id, file_path, start_line, end_line, content, vector, hash, updated_at,
				1 - (vector <=> $1) as score
			FROM chunks
			WHERE project_id = $2 AND file_path LIKE 'path_to_project/%'
			ORDER BY vector <=> $1
			LIMIT $3
```

I notice 2 things:
1) The query runs MUCH faster - 1.5 seconds instead of 30+ seconds.
2) I get 10 results from the correct project.

To test my theory I tried increasing my --limit clause to 500:
```
jeremy@pop-os-thelio:~/tymemud$ grepai search --workspace tymemud --project spell_restore_agent "return" --limit 500
Found 3 results for: "return" in workspace "tymemud"
```

Why on EARTH would you not push this filter down to the SQL??

### Steps to Reproduce

1. Setup grepai workspace on large workspace where many branches are checked out at the same time across multiple (20+) "project" folders 
2. Add those 20+ project folders as "projects"
3. Run grepai search with a "reasonable" limit clause, like 5-10.
4. Notice no results
5. Run grepai search on oversized limit clause, like 500
6. Notice few results.

### Expected Behavior

I expected grepai to push down the --project filter to the database SQL query 

### Actual Behavior

Instead it tries to pull all results and then filters locally, causing the "--limit" clause to not work as expected and causes VERY SLOW query times.

### grepai Version

grepai version 0.29.0

### Operating System

Linux

### Embedding Provider

Ollama

### Relevant Logs

```shell

```

### Configuration

```yaml

```

### Checklist

jeremyakers:
## Description

Added --path command line filter. Pushes filter to postgres when using postgres backend.

## Related Issue

Fixes #130 

## Type of Change

- [X] New feature (non-breaking change that adds functionality)
- [ ] Documentation update
- [ ] Performance improvement
- [ ] Test update

## How Has This Been Tested?

Tested on my codebase using both GOB and Postgres backends.

- [ ] Unit tests
- [ ] Integration tests
- [X] Manual testing

**Test Configuration:**
- OS: Ubuntu 22.04
- Go version: go version go1.25.6 linux/amd64
- Embedding provider: Ollama

## Checklist

- [x] I have updated the documentation if needed
- [x] I have added an entry to CHANGELOG.md (if applicable)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -46,3 +46,7 @@ __debug*
 # Local tooling
 .pre-commit-config.yaml
 docs/plans/
+
+# Example files
+OPENROUTER.md
+openrouter-example.sh
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,23 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   - Efficiently filters results at the database layer for optimal performance
   - Available in MCP `grepai_search` tool via `path` parameter
 
+## [0.32.1] - 2026-02-19
+
+### Changed
+
+- **Dependencies**: Bump `github.com/mark3labs/mcp-go` from 0.43.2 to 0.44.0 (#145) - @dependabot
+
+## [0.32.0] - 2026-02-19
+
+### Added
+
+- **Synthetic & OpenRouter Providers**: Add Synthetic API and OpenRouter as new embedding providers (#106) - @Revaz-Goguadze
+  - Synthetic API (`synthetic`): Cloud embedding via `https://api.synthetic.new` with `nomic-embed-text-v1.5` (768 dims)
+  - OpenRouter (`openrouter`): Multi-provider gateway via `https://openrouter.ai/api/v1` with model selection (text-embedding-3-small, text-embedding-3-large, qwen3-embedding-8b)
+  - Embedder factory pattern (`NewFromConfig`/`NewFromWorkspaceConfig`) centralizing provider initialization across CLI and MCP server
+  - Interactive model selection for OpenRouter during `grepai init`
+  - `--model` flag for non-interactive OpenRouter configuration
+
 ## [0.31.0] - 2026-02-13
 
 ### Added
@@ -564,7 +581,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Initial public release
 
-[Unreleased]: https://github.com/yoanbernabeu/grepai/compare/v0.31.0...HEAD
+[Unreleased]: https://github.com/yoanbernabeu/grepai/compare/v0.32.1...HEAD
+[0.32.1]: https://github.com/yoanbernabeu/grepai/compare/v0.32.0...v0.32.1
+[0.32.0]: https://github.com/yoanbernabeu/grepai/compare/v0.31.0...v0.32.0
 [0.31.0]: https://github.com/yoanbernabeu/grepai/compare/v0.30.0...v0.31.0
 [0.30.0]: https://github.com/yoanbernabeu/grepai/compare/v0.29.0...v0.30.0
 [0.29.0]: https://github.com/yoanbernabeu/grepai/compare/v0.28.0...v0.29.0
--- a/cli/init.go
+++ b/cli/init.go
@@ -14,6 +14,7 @@ import (
 
 var (
 	initProvider       string
+	initModel          string
 	initBackend        string
 	initNonInteractive bool
 	initInherit        bool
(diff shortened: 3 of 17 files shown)


yoanbernabeu:
Hey @jeremyakers, thanks for this contribution! The `--path` filter is a great addition, and the overall approach is solid (especially the SQL push-down for Postgres). A few things I'd like to see addressed before merging:

**CI is currently failing**

- **Lint**: failing (likely related to the `.golangci.yml` changes)
- **Test (ubuntu-latest, Go 1.22)**: failing

**1. `VectorStore` interface: breaking change**

Adding `pathPrefix` directly to the `Search()` signature is a breaking change for anyone implementing this interface. This will happen again every time we add a new filter. I'd suggest introducing a `SearchOptions` struct instead:

```go
type SearchOptions struct {
    PathPrefix string
}

Search(ctx context.Context, queryVector []float32, limit int, opts SearchOptions) ([]SearchResult, error)
```

This way, future filters can be added without breaking the interface again.

**2. Unrelated changes should be in separate PRs**

This PR bundles several changes that aren't related to the `--path` feature:
- `.golangci.yml` version bump and format change (likely causing the lint failure)
- Removal of `runWatchLoop` and `isBackgroundChild` parameter
- `.gitignore` additions for dev notes (`PATH_PREFIX_FEATURE.md`, `IMPLEMENTATION_SUMMARY.md`, etc.)

Could you split these out? It makes the PR easier to review and the git history cleaner.

jeremyakers:
Opened a new PR #141 with just the --path changes.

I'm not sure I would all them "unrelated" though -- The MCP discovery changes are needed for an agent to determine what "path" to provide, since the paths must be relative to the root of the project or workspace. Without providing that information the --path option is not really usable to any agents.

yoanbernabeu:
Hey @jeremyakers, thanks for this great contribution! The `--path` filter feature looks solid and well-structured.

However, the **lint check is currently failing** on CI, which also blocks the build step. Could you please fix the lint errors so we can proceed with the merge?

You can run locally:
```bash
make lint
```

Once the lint issues are resolved and CI is green, we'll be happy to merge this! 🙏

jeremyakers:
Hi @yoanbernabeu - I guess my earlier comment wasn't clear: Based on your earlier feedback about splitting up the unrelated changes, I abandoned this PR (#131) and instead opened two new PRs:
#141 with just the --path changes
and #144 with the MCP discovery changes that allow an agent to effectively find the correct paths to use with the path filter via the MCP server.

So the PR you're currently commenting on (#131) should be closed without merging, and instead merge #141 and #144 instead.

Which btw, I see you did already merge #141 - but #144 is also needed to make the path filter work with agents via MCP.

Hope that makes sense.

jeremyakers:
Actually it looks like I can close this myself. So I am closing this PR as I replaced this PR with #141 and #144 instead.

PR #142, opened 2026-02-15
fix: worktree dedup and multi-project chunk storage

justinkatzman:
## Problem

Two bugs break content-hash dedup when multiple projects (e.g., git worktrees) share a PostgreSQL backend.

### Bug 1: Files with documents but no chunks are permanently skipped

When a prior indexing run creates document records but fails to generate embeddings (e.g., missing API key, transient OpenAI error), subsequent runs skip those files forever. The mod-time gate and hash check in `IndexAllWithBatchProgress` don't verify that chunks actually exist — they assume if a document record is present, the file was fully indexed.

### Bug 2: Chunk ID collision across projects

Chunk IDs use relative paths (e.g., `src/App.tsx_0`). The `chunks` table has `PRIMARY KEY (id)` — not scoped by `project_id`. When a second project inserts chunks with the same relative paths, `ON CONFLICT (id) DO UPDATE` fires and overwrites the first project's rows. But `project_id` is not in the UPDATE set, so it retains the original project's ID. The second project ends up with 0 chunks under its own `project_id`.

This defeats the content-hash dedup feature entirely for worktrees — the cache lookup finds vectors from the first project, `SaveChunks` appears to succeed, but the upsert just overwrites existing rows without changing ownership.

### `indexer/indexer.go`
- Hoist `GetDocument` above the mod-time gate (single call per file instead of two)
- Add `len(doc.ChunkIDs) > 0` to the mod-time gate and hash check
- Add `len(doc.ChunkIDs) == 0` check to `NeedsReindex()` for the watcher path

### `store/postgres.go`
- Auto-migrate `PRIMARY KEY (id)` → `PRIMARY KEY (project_id, id)` in `ensureSchema`
- Update `ON CONFLICT (id)` → `ON CONFLICT (project_id, id)` in `SaveChunks`
- Early return for empty chunks slice

The schema migration is idempotent — it only runs when the existing PK has a single column.

## Testing

Tested with 3 projects sharing one PostgreSQL instance (main repo + 2 worktrees, ~1600 files each):

| Metric | Before | After |
|--------|--------|-------|
| Worktree chunks | 0 (silently lost) | 10,371 |
| Cache hit rate | 0% | 100% |
| Index time per worktree | ~5 min | ~9 sec |
| OpenAI API cost per worktree | ~$0.10 | ~$0.007 |

### DIFF
--- a/cli/watch_initialscan_test.go
+++ b/cli/watch_initialscan_test.go
@@ -142,6 +142,16 @@ func TestRunInitialScan_SkipsIndexedFileByLastIndexTime(t *testing.T) {
 	vecStore := store.NewGOBStore(filepath.Join(projectRoot, "index.gob"))
 	idx := indexer.NewIndexer(projectRoot, vecStore, emb, chunker, scanner, time.Now().Add(1*time.Hour))
 
+	// Seed a document with ChunkIDs so the lastIndexTime gate can skip it.
+	// The new logic requires doc != nil && len(doc.ChunkIDs) > 0 to skip.
+	if err := vecStore.SaveDocument(ctx, store.Document{
+		Path:     "main.go",
+		Hash:     "seeded",
+		ChunkIDs: []string{"c1"},
+	}); err != nil {
+		t.Fatalf("failed to seed document: %v", err)
+	}
+
 	symbolStore := trace.NewGOBSymbolStore(filepath.Join(projectRoot, "symbols.gob"))
 	defer symbolStore.Close()
 
@@ -210,9 +220,10 @@ func TestHandleFileEvent_SkipsUnchangedFile(t *testing.T) {
 		t.Fatalf("failed to scan source file: %v", err)
 	}
 	if err := vecStore.SaveDocument(ctx, store.Document{
-		Path:    "main.go",
-		Hash:    fileInfo.Hash,
-		ModTime: time.Unix(fileInfo.ModTime, 0),
+		Path:     "main.go",
+		Hash:     fileInfo.Hash,
+		ModTime:  time.Unix(fileInfo.ModTime, 0),
+		ChunkIDs: []string{"c1"},
 	}); err != nil {
 		t.Fatalf("failed to seed document: %v", err)
 	}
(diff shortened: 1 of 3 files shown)


yoanbernabeu:
## PR Review

Great PR with an excellent description — the problem analysis, fix strategy, and testing metrics are really well documented. The code changes look correct and well-targeted for the PostgreSQL multi-project scenario.

### What looks good

- The composite PK migration (`project_id, id`) is idempotent and correct
- The `ON CONFLICT` clause update is consistent with the new PK
- The `len(doc.ChunkIDs) > 0` guards correctly handle the "document exists but chunks failed" edge case
- The `SaveChunks` early return for empty slices is a good defensive addition
- The `NeedsReindex` fix is consistent with the indexer changes
- **Manual GOB backend testing passed**: indexing, re-indexing (skip behavior), and search results are identical between `main` and this branch

### Failing tests (5 failures)

However, this PR introduces **5 test regressions** that need to be fixed before merging:

**`indexer` package (2 failures):**

1. **`TestIndexAllWithProgress_BranchSwitchSkipsBulkWithoutLookupOrEmbedding`** (`indexer_branchswitch_test.go:50`)
   - Expected 0 indexed files, got 200
   - Root cause: Moving `GetDocument()` before the mod-time gate means when `doc == nil` (empty mock store), the condition `doc != nil && len(doc.ChunkIDs) > 0` is `false`, so the mod-time filter is bypassed entirely

2. **`TestIndexAllWithProgress_UnchangedFilesSkipped`** (`indexer_test.go:245`)
   - `GetDocument should not be called for files with matching ModTime`
   - Root cause: `GetDocument` is now called for ALL files regardless of mod-time, breaking the assumption that unchanged files skip the store lookup

**`cli` package (3 failures):**

3. **`TestRunInitialScan_SkipsIndexedFileByLastIndexTime`** (`watch_initialscan_test.go:183`)
4. **`TestHandleFileEvent_SkipsUnchangedFile`** (`watch_initialscan_test.go:253`)
5. **`TestHandleWorkspaceFileEvent_SkipsUnchangedFile`** (`watch_initialscan_test.go:336`)
   - All three: embedding is called when it shouldn't be, because `NeedsReindex` now returns `true` for documents with `len(doc.ChunkIDs) == 0` — which is the case in the mock setups

### Suggested approach

The tests need to be updated to reflect the new behavior while preserving the performance optimization for the common case. A possible approach:

- For tests where the mock store has documents: ensure `ChunkIDs` is populated in the mock data
- For the branch-switch test (empty store): consider whether `doc == nil` should still respect the mod-time gate (perhaps `!idx.lastIndexTime.IsZero() && (doc == nil || len(doc.ChunkIDs) > 0)`)
- For `NeedsReindex` tests: ensure mock documents include `ChunkIDs`

All other tests pass (including all store, search, config, embedder, trace, RPG, MCP, daemon, and git packages).

PR #144, opened 2026-02-15
feat(search): Add MCP Discovery commands for workspaces and projects to expose relative paths to use for searching

### LINKED ISSUE #130: [Bug]: Workspaces with projects are Completely Broken - doesn't pass project filter to SQL

jeremyakers:
## Description

Adds MCP discovery tools for listing workspaces and projects to an agent:

- Adds grepai_list_workspaces to discover available workspace names.
- Adds grepai_list_projects to list projects for a selected workspace.
- Adds regression tests

Since path filtering is relative to the root of the workspace or project being searched, in order for an agent to know what "path" to provide they need to know the base bath that a project or workspace is at. This allows agents to see the relative paths for the workspaces/projects so they can provide the correct path to search.

## Related Issue

Follow-up to maintainer feedback on PR #131  (split out unrelated MCP discovery work into its own PR).
Partially Fixes #130 

## Type of Change

- [x] New feature (non-breaking change that adds functionality)
- [ ] Documentation update
- [ ] Performance improvement
- [ ] Test update

## How Has This Been Tested?

- [x] Unit tests
- [x] Integration tests
- [x] Manual testing

**Test Configuration:**
- OS: Ubuntu 22.04
- Go version: go1.25.6
- Embedding provider: Ollama

## Checklist

- [x] I have updated the documentation if needed
- [x] I have added an entry to CHANGELOG.md (if applicable)

## Screenshots (if applicable)

N/A

## Additional Notes

This is a follow-up to maintainer feedback on PR #131  (split out unrelated MCP discovery work into its own PR).

### DIFF
--- a/mcp/server.go
+++ b/mcp/server.go
@@ -590,6 +590,29 @@ func (s *Server) handleWorkspaceSearch(ctx context.Context, query string, limit 
 		results = filteredResults
 	}
 
+	if normalizedPath != "" && len(results) == 0 {
+		selected := resolvedProjects
+		if len(selected) == 0 {
+			selected = listWorkspaceProjectNames(ws.Projects)
+		}
+		projects := selectWorkspaceProjects(ws, selected)
+
+		hasIndexedMatch, matchErr := workspacePathHasIndexedFiles(ctx, st, ws.Name, selected, normalizedPath)
+		if matchErr != nil {
+			log.Printf("Warning: failed to inspect indexed workspace paths for validation: %v", matchErr)
+		} else if !hasIndexedMatch {
+			return mcp.NewToolResultError(
+				buildWorkspacePathValidationError(
+					pathPrefix,
+					selected,
+					workspaceProjectRoots(projects),
+					workspacePathExamples(projects),
+					"no indexed files matched this path prefix in selected projects",
+				),
+			), nil
+		}
+	}
+
 	var data any
 	if compact {
 		searchResultsCompact := make([]SearchResultCompact, len(results))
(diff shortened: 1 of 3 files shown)


yoanbernabeu:
Thanks for the great work on this PR, @jeremyakers! The MCP discovery tools are well-implemented and the code follows the project conventions nicely. Retrocompatibility has been verified — merging! 🚀

PR #181, opened 2026-03-09
Fix OpenAI init model handling and config defaults

garitar:
## Summary
- honor explicit `grepai init --provider openai --model ...` requests instead of always writing the small model
- preserve the documented OpenAI `parallelism: 4` default in generated configs
- infer embedding dimensions from the selected OpenAI/OpenRouter model when dimensions are omitted

## Details
`init` previously hard-coded `text-embedding-3-small` for OpenAI in the non-interactive path, so `--model text-embedding-3-large` was ignored. The config layer also treated omitted dimensions as `1536` for all OpenAI/OpenRouter models, which is incorrect for `text-embedding-3-large` and `qwen3-embedding-8b`.

This patch keeps the upstream default model as small, but fixes three bugs:
- explicit `--model` is respected for OpenAI and OpenRouter init flows
- default OpenAI config objects and generated init configs include `parallelism: 4`
- `GetDimensions()` returns `3072` for `text-embedding-3-large` and `4096` for `qwen3-embedding-8b` when dimensions are omitted

## Validation
- `go test -v -race ./...`
- `go build -o /Users/gary/AI/vendor/grepai/bin/grepai ./cmd/grepai`

One store test remains environment-skipped by the upstream suite when no PostgreSQL DSN is configured.

### DIFF
--- a/cli/init.go
+++ b/cli/init.go
@@ -40,7 +40,7 @@ This command will:
 
 func init() {
 	initCmd.Flags().StringVarP(&initProvider, "provider", "p", "", "Embedding provider (ollama, lmstudio, openai, synthetic, or openrouter)")
-	initCmd.Flags().StringVarP(&initModel, "model", "m", "", "Embedding model (for openrouter: text-embedding-3-small, text-embedding-3-large, qwen3-embedding-8b)")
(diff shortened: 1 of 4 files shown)


yoanbernabeu:
Thanks for the contribution @garitar! Great fix — the model resolution logic is much cleaner now with the centralized `resolveInitModel()` function, and the dimension inference aligns correctly with the actual OpenAI API behavior. Appreciate the thorough tests as well. 🎉

PR #182, opened 2026-03-09
Fix OpenAI workspace create defaults

garitar:
## Summary
- fix OpenAI `workspace create` so generated workspaces write `parallelism: 4` instead of `0`
- reuse shared config defaults for OpenAI/Ollama/LM Studio endpoints and default models in workspace creation paths
- keep the OpenAI workspace default model as `text-embedding-3-small`

## Details
This PR is intentionally separate from #181.

`grepai workspace create` had its own defaulting logic and did not set `embedder.parallelism` for OpenAI, so generated `~/.grepai/workspace.yaml` entries contained `parallelism: 0`. The OpenAI default model remains the current upstream product default (`text-embedding-3-small`); this change only fixes parity and explicitness for workspace config generation.

The patch:
- sets `parallelism` to `4` for OpenAI workspace creation in both the shared flag path and interactive path
- reuses shared config constants where available instead of repeating string literals
- adds focused tests for OpenAI workspace creation defaults

## Validation
- `go test ./cli -run 'TestCreateWorkspaceNonInteractive|TestIntegration_WorkspaceCreateAndMCPResolve'`
- `go build -o /Users/gary/AI/vendor/grepai/bin/grepai ./cmd/grepai`
- isolated runtime check: `grepai workspace create demo --backend postgres --provider openai --yes` now writes `parallelism: 4` while keeping `model: text-embedding-3-small`

### DIFF
--- a/cli/workspace.go
+++ b/cli/workspace.go
@@ -23,6 +23,8 @@ var (
 	workspaceStatusUIRunner   = runWorkspaceStatusUI
 )
 
+const workspaceCreateOpenAIParallelism = 4
+
 var workspaceCmd = &cobra.Command{
 	Use:   "workspace",
 	Short: "Manage multi-project workspaces",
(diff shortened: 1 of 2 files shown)


yoanbernabeu:
Thanks for the contribution @garitar! Clean fix with good test coverage. 🚀

PR #185, opened 2026-03-09
feat(docker): add official Docker image

Greite:
## Summary

- Add a minimal `FROM scratch` Docker image (~22MB) with only the static binary and CA certificates — no Alpine, no tini, no shell
- Add `config.GenerateFromEnv()` to generate `.grepai/config.yaml` from `GREPAI_*` environment variables
- Add `--auto-init` flag to `watch` command to trigger env-based config generation at startup
- Add CI workflow, Docker Compose watch profile, Dependabot config for Docker, and full documentation

## Test plan

- [x] `go test -v -run TestGenerateFromEnv ./config/...` — all 7 tests pass
- [x] `make test` — full test suite passes (0 failures)
- [x] `make build` — binary compiles
- [x] `docker build -t grepai .` — image builds successfully
- [x] `docker images grepai` — verify size ~22MB
- [x] `docker run --rm --entrypoint /grepai grepai version` — binary runs
- [x] Config generation with env vars (see CI tests)
- [x] Config idempotence (existing config not overwritten)

### DIFF
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -46,9 +46,9 @@ jobs:
 
       - name: Test - config not overwritten (idempotent)
         run: |
-          mkdir -p /tmp/grepai-test-idem/.grepai && sudo chown -R 65534:65534 /tmp/grepai-test-idem
+          mkdir -p /tmp/grepai-test-idem/.grepai
           echo -e "version: 1\ncustom: true" > /tmp/grepai-test-idem/.grepai/config.yaml
-          sudo chown 65534:65534 /tmp/grepai-test-idem/.grepai/config.yaml
+          sudo chown -R 65534:65534 /tmp/grepai-test-idem
           docker run --rm \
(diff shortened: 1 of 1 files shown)


yoanbernabeu:
Hi @Greite, thanks for this well-crafted PR — the `FROM scratch` approach, env-based config generation, tests, and documentation are all solid work.

However, after consideration, I've decided not to add official Docker support to grepai. The project is designed to be a lightweight CLI tool, and I want to keep it that way — no container images to maintain, no env-based config path to support alongside the existing `init` flow.

Adding Docker would expand the maintenance surface significantly (CI workflow, GHCR publishing, compose profiles, Dependabot for Docker, etc.) for a tool that's meant to be installed and run directly on the developer's machine.

I appreciate the effort you put into this. If you find Docker useful for your own workflow, feel free to maintain a fork — but this won't be merged upstream.

Thanks again for contributing!

PR #199, opened 2026-03-24
Fix symlink and junction path resolution on all platforms

moabualruz:
## What's wrong

grepai crashes with `"failed to resolve symlinks"` if your terminal opens a project through a symlinked or junctioned parent directory. This is a common scenario on Windows when using NTFS junctions from Dev Drives, OneDrive folder redirects, or a simple `mklink /J workspace D:\work` setup. It also affects macOS and Linux users with symlinked home directories or NFS mounts.

The root cause is that `filepath.EvalSymlinks` fails on certain path types (especially Windows junctions and reparse points), and the code treats that failure as fatal instead of falling back gracefully.

## What this fixes

This PR adds a small helper (`internal/pathutil.ResolveReal`) that tries `EvalSymlinks` first; if it fails, it checks whether the path actually exists before falling back to `filepath.Abs`. This is the same pattern used by golangci-lint ([PR #5245](https://github.com/golangci/golangci-lint/pull/5245)), battle-tested in a project with 18k+ stars.

Every raw `EvalSymlinks` call in the codebase has been replaced with this helper:

| Where | What was happening |
|-------|-------------------|
| `FindProjectRoot` | Hard crash; tool completely unusable |
| `FindProjectRootWithGit` | Crash depending on whether config existed |
| `canonicalPath` (watch) | Silent mismatch; watch events dropped because map keys didn't match |
| `FindWorkspaceForPath` | Target and project paths resolved differently, causing comparison failures |
| `normalizeForPathMatch` (search) | Already had a fallback, but improved with the shared helper |
| `replaceWindowsBinary` (updater) | Crash during self-update |

## Why it works

grepai already stores file paths as relative to the project root. The scanner calls `filepath.Rel(root, path)` and everything downstream uses those relative paths. The only thing that matters is resolving the root consistently; if both indexing and searching resolve to the same root, all relative paths match automatically. No need to double-index or store both path forms.

## What was not changed

- No OS-specific code (`runtime.GOOS` checks or build tags); the fallback is safe on every platform
- No changes to indexing, search, embedding, or MCP logic
- Symlinked subdirectories *inside* a project (e.g., `vendor/` pointing elsewhere) are a separate concern. `WalkDir` does not follow those by design, and fixing that would require cycle detection. Happy to tackle that in a follow-up if there is interest.

## Testing

- Added tests for `ResolveReal` covering real paths, nonexistent paths, and symlink chains
- Fixed `TestFindProjectRootWithSymlink`, which was skipping entirely on Windows. Go 1.22+ creates directory junctions without elevation, so the blanket skip was unnecessary.
- All existing tests pass with `-race`
- Manually verified from a junctioned path (`C:\Users\mkh\workspace\` -> `D:\work\`): `init`, `search`, `watch` all work where they previously crashed

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+
+- **Multi-Model Index Support**: New `store.multi_model` config option tags chunks with their embedding provider/model. When enabled, search returns only chunks matching the current model, enabling safe model switching without full re-indexing.
+- **migrate-model Command**: New `grepai migrate-model <provider/model>` command stamps legacy untagged chunks with a model identifier. Metadata-only operation; no embedding API calls.
+- **Startup Validation**: Watch and search commands now validate that no untagged chunks exist when `multi_model` is enabled, with a clear error message directing users to run `migrate-model`.
+
 ## [0.35.0] - 2026-03-16
 
 ### Added
--- a/cli/migrate_model.go
+++ b/cli/migrate_model.go
@@ -1,0 +1,106 @@
+package cli
+
+import (
+	"context"
+	"fmt"
+	"strings"
+
+	"github.com/spf13/cobra"
+	"github.com/yoanbernabeu/grepai/config"
+	"github.com/yoanbernabeu/grepai/store"
+)
+
+var migrateModelCmd = &cobra.Command{
+	Use:   "migrate-model <provider/model>",
+	Short: "Stamp untagged chunks with an embedding model identifier",
+	Long: `Stamp all chunks that have an empty EmbedModel field with the given
+provider/model string. This is required when enabling multi_model on an
+existing index so that legacy chunks become searchable under the new
+filtering rules.
+
+No embedding API calls are made; this is a metadata-only operation.
+
+Example:
+  grepai migrate-model ollama/nomic-embed-text`,
+	Args: cobra.ExactArgs(1),
+	RunE: runMigrateModel,
+}
+
+func init() {
+	rootCmd.AddCommand(migrateModelCmd)
+}
+
+func runMigrateModel(cmd *cobra.Command, args []string) error {
+	modelTag := args[0]
+
+	// Validate model tag format
+	if !strings.Contains(modelTag, "/") {
+		return fmt.Errorf("model tag must be in provider/model format (e.g. ollama/nomic-embed-text), got %q", modelTag)
+	}
+
+	parts := strings.SplitN(modelTag, "/", 2)
+	if parts[0] == "" || parts[1] == "" {
+		return fmt.Errorf("model tag must have non-empty provider and model (e.g. ollama/nomic-embed-text), got %q", modelTag)
+	}
+
+	ctx := context.Background()
+
+	// Find project root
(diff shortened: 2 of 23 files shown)


yoanbernabeu:
Hi @moabualruz, thanks a lot for this contribution — the symlink/junction fix is genuinely valuable and the `internal/pathutil.ResolveReal` helper (modeled after the golangci-lint pattern) is clean and well tested. I'd love to land that part.

However, I can't merge this PR as it stands because **the scope doesn't match the title/description**. The PR is advertised as a symlink/junction fix, but it also bundles two unrelated features:

1. **Multi-Model Index Support** (`store.multi_model` config, `EmbedModel` chunk tagging, search filtering)
2. **`grepai migrate-model` command** (a brand-new CLI command, ~106 lines)
3. **Startup validation** for `watch` and `search` tied to the multi-model feature

A few concrete issues with the multi-model part:
- The `EmbedModel` filter is implemented **only in `store/gob.go`**. Both `store/postgres.go` and `store/qdrant.go` are untouched, so enabling `multi_model` with Postgres or Qdrant would silently return unfiltered results — contradicting the documentation added in this PR.
- `EmbedModel` is added to `store.Chunk`, which impacts the on-disk GOB format. That kind of change deserves a dedicated discussion/issue before implementation.
- The CHANGELOG entries in this PR only mention the multi-model feature — not the symlink fix that's supposedly the topic of the PR.

Also, CI is currently red across Linux and macOS due to a leftover unused import:
```
config/config_test.go:6:2: "runtime" imported and not used
```

### What I'd like to do

I'm going to **close this PR**, but please don't take that as a rejection of your work — quite the opposite. Could you split it into separate PRs, one per concern?

1. **PR 1 — Symlink/junction fix** (the part I want to merge first):
   - `internal/pathutil/` (helper + tests)
   - All the `EvalSymlinks` → `pathutil.ResolveReal` call sites (config, workspace, watch, search, updater)
   - The `TestFindProjectRootWithSymlink` Windows skip removal
   - Fix the unused `runtime` import so CI is green

2. **PR 2 — Multi-Model Index Support** — please open a GitHub issue first so we can discuss the design (especially the GOB schema change, behavior across all 3 store backends, and the migration UX). Once aligned, the implementation PR will be much smoother.

3. **PR 3 — `migrate-model` command** — naturally pairs with PR 2, can be the same PR or a follow-up.

The symlink fix on its own would be a quick merge once isolated. Thanks again for the careful work and for the detailed PR description on the symlink issue — it really helps reviewers.

yoanbernabeu:
Closing in favor of the split suggested above. Looking forward to the symlink-only PR — that one should be a quick merge. Thanks again @moabualruz!

moabualruz:
As requested, this PR has been split into three focused PRs — all rebased onto the current `main`:

1. **#231 — Symlink/junction fix** (the fast merge) — `internal/pathutil.ResolveReal` helper + all `EvalSymlinks` call sites replaced + Windows symlink test fix + unused `runtime` import removed so CI is green.

2. **#200 — Multi-Model Index Support** — `store.multi_model` config, `EmbedModel` field on chunks, search filtering, startup validation. Updated to remove the migrate-model command (now separate) and to fix the `runtime` import CI failure.

3. **#232 — `migrate-model` command** — metadata-only command to stamp legacy chunks; depends on #200.

Thanks for the detailed review feedback — the split makes the history and review scope much cleaner.

PR #200, opened 2026-03-24
feat(store): add embedding model tagging with multi-model index support

moabualruz:
## Summary

Adds optional model tagging on indexed chunks, enabling multi-model coexistence
within a single index. Off by default; zero behavioral change for existing users.

Design discussion and open questions: **#233**

## Problem

When users change their embedding provider or model, grepai silently returns
meaningless search results. Vectors from different embedding spaces cannot be
compared, but grepai has no way to detect or prevent this.

## What changed

- **EmbedModel field on Chunk**: stores `"provider/model"` identifying which
  embedder produced the vector. Only set when `multi_model` is enabled.

- **multi_model config option** (default: false): when enabled, every chunk is
  tagged during indexing and search filters strictly by the current model.
  Chunks without a tag are excluded from filtered searches.

- **Startup validation**: when `multi_model` is true and untagged chunks exist,
  grepai stops with a message directing the user to run `migrate-model` first.

- **Hybrid search**: text search path also respects the model filter when set.

## Design decisions

- `multi_model: false` means zero change. No tags saved, no filtering, identical
  to current behavior. This is the most important invariant.

- `multi_model: true` is strict. Empty-tag chunks are excluded, not included.
  This prevents mixing vectors from unknown models with tagged ones.

- No auto-detection or auto-backfill of model names. The user knows what model
  built their index; they provide it explicitly via `migrate-model`. The tool
  does not guess.

- Model switch with multi_model enabled: new chunks get the new tag, old chunks
  retain theirs. Search returns only the current model's chunks. Switching back
  returns the previous model's chunks without re-indexing.

## Known limitations (tracked in #233)

- Filtering is implemented for GOB backend only. Postgres and Qdrant return unfiltered results if `multi_model` is enabled — startup validation blocks this with an explicit error for non-GOB backends.
- Stale chunks from old models accumulate over time; no cleanup command yet.

## Test plan

- [x] Filter returns only chunks matching the configured model
- [x] Empty-tag chunks excluded when filter is active
- [x] No filter returns all chunks (default behavior)
- [x] multi_model defaults to false
- [x] EmbedModelTag guards against empty provider/model
- [x] Hybrid search text path respects model filter
- [x] Full test suite passes with zero regressions
- [x] Build succeeds

## Related

Split from #199 (closed). Part 2 of 3:
1. #231 — symlink/junction fix
2. This PR — multi-model index support
3. #232 — migrate-model command (depends on this PR)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+
+- **Multi-Model Index Support**: New `store.multi_model` config option tags chunks with their embedding provider/model. When enabled, search returns only chunks matching the current model, enabling safe model switching without full re-indexing.
+- **migrate-model Command**: New `grepai migrate-model <provider/model>` command stamps legacy untagged chunks with a model identifier. Metadata-only operation; no embedding API calls.
+- **Startup Validation**: Watch and search commands now validate that no untagged chunks exist when `multi_model` is enabled, with a clear error message directing users to run `migrate-model`.
+
 ## [0.35.0] - 2026-03-16
 
 ### Added
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+
+- **Multi-Model Index Support**: New `store.multi_model` config option tags chunks with their embedding provider/model. When enabled, search returns only chunks matching the current model, enabling safe model switching without full re-indexing.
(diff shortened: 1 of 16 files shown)


yoanbernabeu:
Hi @moabualruz, this branch currently has merge conflicts with `main` (`mergeStateStatus: DIRTY`). Could you rebase `feat/model-metadata-in-index` onto the latest `main` and resolve the conflicts? Once it's clean and CI is green, I'll review. Thanks!

moabualruz:
@yoanbernabeu Thanks a lot for taking a look, I resolved the conflicts

moabualruz:
Design discussion opened in #233. Feedback on backend coverage (question 2) and stale chunk cleanup (question 4) particularly welcome before merge.

moabualruz:
Closing as superseded by #232, which contains this branch's commits plus the follow-up migration work. Keeping the work visible there avoids two stale overlapping PRs.

PR #196, opened 2026-03-21
fix(worktree): make watch fanout explicit, refresh MCP startup, and paginate Qdrant stats

garitar:
## Summary

This PR tightens several operator-facing behaviors:

- make linked-worktree fanout explicit instead of automatic for `watch`
- treat missing local `.grepai/config.yaml` as uninitialized in linked worktrees
- refresh only the local project context at `mcp-serve` startup instead of broad startup-side indexing behavior
- strengthen direct coverage for MCP serve, background watch argv propagation, and Qdrant stats/file-list pagination
- fix Qdrant stats and file-list paths so they paginate through large collections

## Why

Recent reports point at two related pain areas:

- nested linked worktrees can behave as if they are initialized even when local config is missing: [Issue #189](https://github.com/yoanbernabeu/grepai/issues/189)
- `mcp-serve --workspace` needs a more deliberate startup model that avoids broad repeated indexing and instead favors local-context incremental refresh: [Issue #193](https://github.com/yoanbernabeu/grepai/issues/193)

While tightening coverage around those paths, I also found and fixed a separate Qdrant pagination bug that affects operator-facing stats/file listings on larger collections: [Issue #195](https://github.com/yoanbernabeu/grepai/issues/195)

This change keeps full-current-file indexing semantics, but reduces surprise fanout and tightens startup behavior around the actual local project context.

## Scope

Included:
- explicit opt-in for all-worktree watch fanout
- stricter linked-worktree initialization detection
- local startup refresh path for MCP serve
- direct tests for the touched operator paths
- Qdrant stats/file listing pagination fix discovered while strengthening coverage

Not included:
- embedding-model invalidation: [Issue #139](https://github.com/yoanbernabeu/grepai/issues/139)
- broader subprocess-style orchestration UX: [Issue #194](https://github.com/yoanbernabeu/grepai/issues/194)

## Notes

This is adjacent to, but not a claimed full fix for, the nested-worktree status/accounting report in [Issue #192](https://github.com/yoanbernabeu/grepai/issues/192).

## Validation

- `go test ./cli ./store`
- `make pre-commit`

PR #205, opened 2026-03-29
fix(qdrant): enable incremental indexing via doc_hash on chunk points

drizzt:
## Summary

- Qdrant backend now tracks document hashes (`doc_hash`) as payload on chunk points, enabling `grepai watch` to skip unchanged files on restart instead of re-embedding the entire codebase
- Adds `file_path` field index for efficient filtered lookups used by `SetPayload` and `GetDocument`
- Switches `GetAllChunks` and `ListDocuments` to paginated `scrollAll` for correct behavior on large collections
- Legacy indexes (chunks without `doc_hash`) are migrated transparently on first run via a one-time re-index

## Design

Instead of a separate document collection, `doc_hash` is stored directly on existing chunk points via `SetPayload`. `GetDocument` reads it from any single chunk of the file and returns a sentinel `ChunkIDs` slice to avoid expensive full-file scrolls. This keeps the Qdrant schema flat and avoids cross-collection coordination.

## Test plan

- [x] `TestQdrantStore_GetDocument_WithDocHash` — verifies hash and sentinel ChunkIDs are returned
- [x] `TestQdrantStore_GetDocument_LegacyChunkNoDocHash` — verifies legacy chunks trigger re-index
- [x] `TestQdrantStore_GetDocument_NotFound` — verifies nil return for missing files
- [x] `TestQdrantStore_SaveDocument_SetsPayload` — verifies SetPayload is called with correct doc_hash
- [x] `TestQdrantStore_SaveDocument_EmptyHashIsNoop` — verifies no-op for empty hash
- [x] `TestQdrantStore_ListDocuments_Paginates` — verifies pagination via scrollAll
- [x] Full test suite passes (`go test ./...`)

## Note
This is on top of #196

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,20 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+
+- **Explicit linked-worktree watch opt-in**: `grepai watch --all-worktrees` now enables linked worktree fanout only when requested, keeping default watch sessions scoped to the current worktree.
+
+### Fixed
+
+- **MCP workspace startup refresh**: `grepai mcp-serve --workspace <name>` now keeps full-workspace indexing explicit while refreshing only the local project context when started inside a workspace project.
+- **Linked worktree auto-init detection**: linked worktree watch discovery now treats missing `.grepai/config.yaml` as uninitialized, even when a partial `.grepai/` directory already exists.
+- **Qdrant file counts**: `status` and MCP index status now report file counts correctly for Qdrant-backed indexes instead of always showing `0`.
+
+### Documentation
+
+- Update git worktree and workspace/MCP docs to describe the new watch opt-in and MCP startup behavior.
+
 ## [0.35.0] - 2026-03-16
 
 ### Added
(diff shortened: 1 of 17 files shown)


yoanbernabeu:
Thanks for this, and sorry it sat for so long.

This branch dates from late March and has never been rebased, so it no longer merges. `git merge origin/main` conflicts on two files:

- `CHANGELOG.md` — trivial
- `cli/watch.go` — **not** trivial. `main` has since refactored `initializeWorkspaceRuntime`, extracting `newWorkspaceProjectRuntime` and `runWorkspaceProjectInitialScan`. This branch still builds the `workspaceProjectRuntime` struct inline the old way, so resolving it means re-implanting the `doc_hash` logic into the new helper-based flow rather than picking a side.

The branch is also well behind `main` in general (the Vue/Svelte/Astro framework processors, `search/dedup.go`, `cli/refs.go` and others have landed since).

I'd rather not resolve this myself: getting it wrong risks quietly breaking the recent `watch.go` refactor, and the Qdrant incremental path needs a real retest after the rebase — which is much easier for you, since you have the Qdrant setup that motivated the PR.

Could you rebase onto current `main` and re-verify the incremental indexing flow? The underlying feature (incremental indexing via `doc_hash` on chunk points instead of full re-upload) is valuable and I'd like to take it. Note that #207 has since landed and also touches `store/qdrant.go` (gRPC max message size raised to 64 MB, vectors dropped from `GetAllChunks`) — shouldn't conflict with your changes, but worth knowing about when you retest.

PR #206, opened 2026-03-29
fix(rpg): add Graph mutex to prevent concurrent map access crash

drizzt:
## Summary

- Adds `sync.RWMutex` to the RPG `Graph` struct to prevent concurrent map access crashes
- The crash (`concurrent map iteration and map write`) occurred when the persist goroutine (GOB-encoding `graph.Nodes` every 1s) ran concurrently with the event handler goroutine (calling `RemoveNode`, which deletes from the same map)
- `RemoveEdgesIf` rewritten with snapshot+apply pattern to avoid re-entrant locking (the predicate calls `NodePath` which acquires a read lock; evaluating it while holding a write lock would deadlock)
- `persistUnlocked` snapshots Nodes/Edges under `RLock` before encoding, so graph mutations can proceed during the slow GOB encode
- `BuildFull` uses a new atomic `Reset()` method instead of direct field assignment

## Why only Qdrant triggered this

The race exists for all backends, but Qdrant makes it reliably triggerable:

- **GOB store**: `DeleteByFile` is an in-memory map delete (~microseconds). A burst of events finishes in <1ms — the 1s persist timer never fires during the mutation window.
- **Qdrant**: `DeleteByFile` is a network round-trip (~300–3800ms per event). A burst of 10 events spans multiple persist intervals, giving the timer many chances to fire mid-mutation.

The incremental indexing PR (#205) amplified this by making the initial scan complete in ~10s instead of minutes, pushing the watch loop into steady state much sooner.

## Test plan

- [x] `make test` passes with `-race` flag
- [x] `grepai watch` with Qdrant backend: trigger file changes during watch, confirm no `concurrent map iteration and map write` crash

### DIFF
--- a/rpg/indexer.go
+++ b/rpg/indexer.go
@@ -82,9 +82,7 @@ func (idx *RPGEncoder) BuildFull(ctx context.Context, symbolStore trace.SymbolSt
 	graph := idx.store.GetGraph()
 
 	// Clear existing graph data in-place (not reassigning the pointer)
-	graph.Nodes = make(map[string]*Node)
-	graph.Edges = make([]*Edge, 0)
-	graph.RebuildIndexes()
+	graph.Reset()
 
 	// Get all documents from vector store to know which files to process
 	docs, err := vectorStore.ListDocuments(ctx)
--- a/rpg/model.go
+++ b/rpg/model.go
@@ -3,6 +3,7 @@ package rpg
 import (
 	"fmt"
 	"strings"
+	"sync"
 	"time"
 )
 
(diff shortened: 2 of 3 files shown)


unifirer:
This PR is the right fix for a crash that is still live on `main` and in the latest release (v0.35.0). Posting here to keep it visible / link the duplicate report.

**Verified still unfixed on `main` (today):** `rpg/model.go` `Graph` struct has **no `sync.RWMutex`** — `Nodes`, `Edges`, `byKind`, `byFile`, `adjForward`, `adjReverse` etc. are mutated directly by `AddNode`/`RemoveNode`/`AddEdge`/`RemoveEdgesIf` with zero locking. So the race in the PR description persists.

**Real-world report:** issue #279 — `grepai watch` (v0.35.0, **GOB backend**, `rpg.enabled: true`, `watch.rpg_persist_interval_ms: 1000`) dies after hours with:
```
fatal error: concurrent map iteration and map write
... encoding/gob.(*Encoder).encodeMap(...)  <- persist goroutine (GOB-encodes Graph.Nodes every 1s)
... rpg.(*Graph).RemoveNode(...)           <- event-handler goroutine (delete/rename)
```

Two things to flag for this PR to actually close it:
1. **Rebase against current `main`** — the tree has moved since Mar 29 (the RPG layer was later refactored to `RPGEncoder`).
2. **Cover the GOB persist path, not just Qdrant.** The PR description says only Qdrant reliably triggered the race, but issue #279 shows a reproducible **GOB-backend** crash from the same `gob.Encoder.encodeMap` vs `RemoveNode` collision. `persistUnlocked` snapshots under `RLock` before encoding — that snapshot must wrap the GOB encode too.

Happy to test a rebased build if that helps move it forward.

PR #215, opened 2026-04-05
feat(indexer): use per-file ModTime to skip unchanged files on watch restart

### LINKED ISSUE #216: feat(indexer): use per-file ModTime to skip unchanged files on watch restart
## Problem

Every time `grepai watch` is restarted, it re-scans all files. The existing skip mechanism uses a single global `lastIndexTime` timestamp — any file whose ModTime is before this global cutoff gets skipped. This has two issues:

1. **Imprecise**: A file modified during a long initial scan (between its scan and when `lastIndexTime` is set at the end) could be silently skipped on the next restart
2. **Missed optimization**: The `Document` struct already stores per-file `ModTime`, but this value was never compared against the current file's ModTime for the skip decision

## Proposed Solution

- **Primary check — per-file ModTime**: Compare each file's current ModTime against `doc.ModTime` stored in the index. If they match and the document has chunks, skip immediately (no file I/O needed).
- **Fallback — global `lastIndexTime`**: Preserved for backward compatibility with legacy index files where `doc.ModTime` is zero.
- **ModTime refresh on hash match**: When a file's content hash matches but ModTime differs (e.g., after `git checkout`), update the stored ModTime so future restarts benefit from the fast per-file path.
- **Symbol extraction optimization**: Use the indexer's `UnchangedFiles` set instead of global timestamp comparison for more accurate skip behavior.

## Expected Impact

- Eliminates unnecessary LLM embedding calls for unchanged files on watch restart
- Dramatically reduces startup time on subsequent `grepai watch` runs
- Backward compatible with existing GOB/Postgres/Qdrant stores

iambaljeet:
Closes #216

## Summary

Replace the coarse global `lastIndexTime` check with precise **per-file ModTime comparison** when the watch command restarts. This eliminates unnecessary LLM embedding calls for unchanged files, dramatically reducing startup time on subsequent `grepai watch` runs.

## Problem

Every time `grepai watch` is restarted, it re-scans all files. The existing skip mechanism uses a single global `lastIndexTime` timestamp — any file whose ModTime is before this global cutoff gets skipped. This has two issues:

1. **Imprecise**: A file modified during a long initial scan (between its scan and when `lastIndexTime` is set at the end) could be silently skipped on the next restart
2. **Missed optimization**: The `Document` struct already stores per-file `ModTime`, but this value was never compared against the current file's ModTime for the skip decision

## Solution

**Primary check — per-file ModTime** (lines 144–150 in indexer.go):
- Compare each file's current ModTime against `doc.ModTime` stored in the index
- If they match and the document has chunks → skip immediately (no file I/O needed)

**Fallback — global `lastIndexTime`** (lines 153–159 in indexer.go):
- Preserved for backward compatibility with legacy index files where `doc.ModTime` is zero
- Only used when `doc.ModTime.IsZero()` (old documents created before this change)

**ModTime refresh on hash match** (lines 178–185 in indexer.go):
- When a file's content hash matches but ModTime differs (e.g., after `git checkout`), the stored ModTime is updated
- Future restarts benefit from the fast per-file ModTime path

**Symbol extraction optimization** (line 819 in watch.go):
- Uses the indexer's `UnchangedFiles` set instead of global timestamp comparison
- More accurate — skips only files the indexer actually verified as unchanged

## Backward Compatibility

- **GOB store**: `gob` decoding sets missing `time.Time` fields to zero value; the code gates on `!doc.ModTime.IsZero()` before using the per-file path → legacy files fall back to global `lastIndexTime`
- **Postgres**: `mod_time TIMESTAMP` column already exists; `ON CONFLICT DO UPDATE` handles the new writes
- **Qdrant**: `SaveDocument` is a no-op and `GetDocument` returns empty `ChunkIDs`, so the new per-file gate is never entered

## Test Results

- All 15 test packages pass with `-race` flag
- `go vet ./...` clean
- End-to-end verified with LM Studio (`text-embedding-nomic-embed-text-v1.5`):
  - First run: 3 files indexed (315ms)
  - Restart unchanged: **0 indexed, 3 skipped (0s — no LLM calls)**
  - 1 modified + 1 new: 2 indexed, 2 skipped (158ms)
  - 1 deleted: 1 removed, 3 skipped (0s)

## Checklist

- [x] Tests pass (`go test -race ./...`)
- [x] `go vet ./...` clean
- [x] Backward compatible with existing GOB/Postgres/Qdrant stores
- [x] No new dependencies
- [x] End-to-end tested with real LLM embedder

### DIFF
--- a/cli/watch.go
+++ b/cli/watch.go
@@ -815,12 +815,9 @@ func runInitialScan(ctx context.Context, idx *indexer.Indexer, scanner *indexer.
 			continue
 		}
 
-		// Skip files that are unchanged since the last index run and already tracked.
-		if !lastIndexTime.IsZero() {
-			fileModTime := time.Unix(file.ModTime, 0)
-			if (fileModTime.Before(lastIndexTime) || fileModTime.Equal(lastIndexTime)) && symbolStore.IsFileIndexed(file.Path) {
-				continue
-			}
+		// Skip files verified unchanged by the indexer (per-file ModTime or hash match).
+		if stats.UnchangedFiles[file.Path] && symbolStore.IsFileIndexed(file.Path) {
+			continue
 		}
 
 		fileInfo, err := scanner.ScanFile(file.Path)
--- a/indexer/indexer.go
+++ b/indexer/indexer.go
@@ -22,12 +22,13 @@ type Indexer struct {
 }
 
 type IndexStats struct {
-	FilesIndexed  int
-	FilesSkipped  int
-	ChunksCreated int
-	FilesRemoved  int
-	Duration      time.Duration
-	ScannedFiles  []FileMeta // All files found during scan (for reuse by callers)
+	FilesIndexed   int
+	FilesSkipped   int
+	ChunksCreated  int
+	FilesRemoved   int
+	Duration       time.Duration
+	ScannedFiles   []FileMeta      // All files found during scan (for reuse by callers)
+	UnchangedFiles map[string]bool // Files verified unchanged (by ModTime or hash) for callers to reuse
 }
 
 // ProgressInfo contains progress information for indexing
@@ -93,7 +94,9 @@ func (idx *Indexer) IndexAllWithProgress(ctx context.Context, onProgress Progres
 // When the embedder implements BatchEmbedder, files are processed in parallel using cross-file batching.
 func (idx *Indexer) IndexAllWithBatchProgress(ctx context.Context, onProgress ProgressCallback, onBatchProgress BatchProgressCallback) (*IndexStats, error) {
 	start := time.Now()
-	stats := &IndexStats{}
+	stats := &IndexStats{
+		UnchangedFiles: make(map[string]bool),
+	}
 
 	// Scan all files (metadata-only first pass)
 	fileMetas, skipped, err := idx.scanner.ScanMetadata()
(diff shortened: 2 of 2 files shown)


yoanbernabeu:
Thanks for this — you identified a real bug. The global `lastIndexTime` cutoff genuinely can skip a file that was modified *during* a long initial scan, and per-file ModTime is the right instinct for fixing it.

I can't merge it as it stands, though: it introduces a fast path that can silently drop real edits, and I reproduced it end-to-end.

## What I measured

1. Indexed a small project (2 files).
2. Modified a file's **contents**, then set its mtime to the same second already stored in the index (`touch -t`) — i.e. two edits landing inside the same second.
3. Restarted the watcher:
   ```
   Initial scan complete: 0 files indexed, 0 chunks created, 0 files removed, 2 skipped
   ```
4. `grepai search` for a unique string added in step 2: **not found**. The edit is absent from the index, with no error and no warning.

## Why

`ModTime` is persisted as Unix **seconds** (`info.ModTime().Unix()`), so sub-second precision is discarded even on filesystems that have it. The new code then does:

```go
if fileModTime.Equal(doc.ModTime) {
    continue    // skip — and never falls through to the content hash
}
```

The previous code used mtime only as a coarse pre-filter and **always** hashed the content before deciding. Removing that fallback is what makes the failure silent: one-second granularity plus no hash check means any two edits within the same second collapse into "unchanged".

This matters more than usual for grepai specifically: a large part of the user base is AI agents editing files in rapid succession, which is exactly the pattern that lands multiple writes in the same second.

## Suggested direction

Either of these would work:

- Keep per-file ModTime as a *fast path only*, and fall through to the content hash whenever the mtimes are equal (mismatch → definitely changed, match → still verify). You keep almost all the speed-up, since the hash only runs on the equality case.
- Or persist ModTime with nanosecond precision (`UnixNano`) and keep the hash fallback anyway for filesystems that don't provide it.

A regression test in the spirit of the reproduction above — same mtime, different content, assert the file is re-indexed — would lock this down.

## One more thing

#277 (parallelised change detection) has since landed on `main` and refactored this same loop into a `decideFileScan` helper, so this branch will need a rebase; the per-file ModTime logic wants to live inside that helper now.

Everything else checks out: `go build`, `go vet` and `go test -race ./...` all pass, and the underlying problem you're solving is worth solving. Happy to take another look once a changed file can't slip through.

PR #230, opened 2026-04-22
fix(store): persist GOB index atomically to prevent EOF corruption

Greite:
## Summary

`store.GOBStore.Persist` currently calls `os.Create` on the target path, which truncates `.grepai/index.gob` to zero bytes **before** the gob encoder has written anything. If the process is killed mid-encode, the on-disk index is left truncated and the next `Load` fails with:

```
Error: failed to load index: failed to decode index: unexpected EOF
```

The failure mode exists for OOM kills, host crashes, and power loss.

The fix mirrors the pattern already used by `trace/store.go` (`GOBSymbolStore.persistUnlocked`) and `rpg/store_gob.go`:

1. `os.CreateTemp` a sibling temp file
2. Encode into it
3. `Sync()` (fsync) + `Close()`
4. `fileutil.ReplaceFileAtomically` (rename with cross-device fallback)
5. Deferred cleanup removes the temp file if any step fails

No behavior change in the success path. In the failure path, the previous index stays intact — a crashed persist leaves the last successful snapshot on disk, instead of an empty/truncated file.

## Test plan

- [x] `TestGOBStore_PersistIsAtomic` — new regression test. Holds an open read handle on `index.gob` across a second `Persist`; on POSIX, an atomic rename keeps the previous inode alive for the retained descriptor, so the handle still reads the original bytes. The previous truncate-in-place implementation fails this test (handle sees the second persist's bytes because the same inode was rewritten). Fails on `main`, passes with this change.
- [x] `TestGOBStore_PersistDoesNotLeaveTempFiles` — asserts the cleanup path on success.
- [x] `go test -race ./...` — full suite green, no regressions.
- [x] `go vet ./...` — clean.
- [x] `gofmt -l store/` — clean.

## Notes

- `trace/store.go` and `rpg/store_gob.go` already use this exact pattern, so `store/gob.go` was the only remaining caller of `os.Create` on a persisted index path.
- The helper `fileutil.ReplaceFileAtomically` already existed and is unchanged.
- The old buggy code was technically non-atomic on all platforms; the new code uses POSIX rename semantics on Unix and Windows's `MoveFileEx` under the hood (with the helper's remove+rename fallback for cross-device cases).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Fixed
+
+- **Atomic Index Persist**: `store.GOBStore.Persist` now writes the vector index via a temp file + `fsync` + atomic rename, mirroring the pattern already used by `trace/store.go` and `rpg/store_gob.go`. Previously the index was truncated in place with `os.Create` before being progressively encoded, so any crash mid-encode (SIGKILL after Docker's `stop_grace_period`, OOM, power loss) corrupted `.grepai/index.gob` and made the next `Load` fail with `failed to decode index: unexpected EOF`.
+
 ## [0.35.0] - 2026-03-16
 
 ### Added
--- a/store/gob.go
+++ b/store/gob.go
@@ -6,6 +6,7 @@ import (
 	"fmt"
 	"math"
 	"os"
+	"path/filepath"
 	"sort"
 	"strings"
 	"sync"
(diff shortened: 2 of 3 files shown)


yoanbernabeu:
Thank you for this fix — the diagnosis was spot on, and the atomic-write pattern you used (temp file → `Sync` → `fileutil.ReplaceFileAtomically`, mirroring `trace/store.go` and `rpg/store_gob.go`) is exactly right.

I'm closing this in favour of #269, which landed the same atomic persist plus a second layer this PR doesn't have: self-healing on the **read** path. `GOBStore.loadUnlocked` now quarantines an already-truncated `index.gob` to `index.gob.corrupt` and starts from an empty index instead of failing forever.

That second layer turned out to matter. I reproduced the failure end-to-end: index built, truncated to simulate a crash mid-write, then reopened.

- On `main` (and with this PR's code, which doesn't touch the decode path): `Error: failed to load index: failed to decode index: unexpected EOF` — on `search`, on `status`, and on `watch`. A user already stuck in that state stays stuck after upgrading, because the atomic write only protects *future* writes.
- With #269: `Warning: index file .../index.gob is corrupted (unexpected EOF); moved it to .../index.gob.corrupt and starting with an empty index — it will be rebuilt on the next scan`, then `Initial scan complete: 4 files indexed, 7 chunks created` and search works again.

Since the users who reported this are by definition already holding a corrupted index, recovering them mattered as much as preventing new corruption — so I took the superset.

This is not a reflection on the quality of your patch: it was correct, well-explained, and had good test coverage. Two contributors independently landed the same root-cause fix, and I had to pick one. Thanks for taking the time on it, and please do send more — the analysis in your description was genuinely useful in confirming the diagnosis.

PR #231, opened 2026-04-23
fix(pathutil): resolve symlinks and junctions gracefully on all platforms

moabualruz:
## Summary

Adds `internal/pathutil.ResolveReal` — a small helper that tries `filepath.EvalSymlinks` first and falls back to `filepath.Abs` when EvalSymlinks fails (e.g. NTFS junctions, reparse points, NFS mounts). Replaces every raw `EvalSymlinks` call in the codebase with this helper.

This is a focused extraction from #199 (closed), which bundled this fix with unrelated multi-model features. Closes the crash scenario described there.

## What was changed

| Call site | Effect |
|-----------|--------|
| `FindProjectRoot` | No more hard crash; falls back to `filepath.Abs` |
| `FindProjectRootWithGit` | Same fallback applied |
| `canonicalPath` (watch) | Consistent map keys; watch events no longer silently dropped |
| `FindWorkspaceForPath` | Target and project paths resolve symmetrically |
| `normalizeForPathMatch` (search) | Shared helper replaces ad-hoc inline fallback |
| `replaceWindowsBinary` (updater) | No crash during self-update |

## Tests

- Added `internal/pathutil/` with unit tests covering real paths, nonexistent paths, and symlink chains
- `TestFindProjectRootWithSymlink`: removed blanket Windows skip — Go 1.22+ creates directory junctions without elevation
- Fixed unused `runtime` import in `config/config_test.go` (was causing CI red in #199)
- All existing tests pass with `-race`

## Related

Split from #199 as requested. The multi-model index support and `migrate-model` command are being submitted as separate PRs.

### DIFF
--- a/cli/watch.go
+++ b/cli/watch.go
@@ -24,6 +24,7 @@ import (
 	"github.com/yoanbernabeu/grepai/git"
 	"github.com/yoanbernabeu/grepai/indexer"
 	"github.com/yoanbernabeu/grepai/rpg"
+	"github.com/yoanbernabeu/grepai/internal/pathutil"
 	"github.com/yoanbernabeu/grepai/store"
 	"github.com/yoanbernabeu/grepai/trace"
 	"github.com/yoanbernabeu/grepai/watcher"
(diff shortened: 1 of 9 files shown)


moabualruz:
Closing from the contributor side to avoid leaving stale open PRs indefinitely. The branch `fix/windows-symlink-resolution` remains the preserved work reference if the upstream maintainer wants to revive or request it later.

PR #248, opened 2026-05-20
Fix Qdrant chunk identity collisions

### LINKED ISSUE #247: [Bug]: Cross-project Qdrant vector index overwrite via non-namespaced chunk IDs
### Bug Description

When multiple grepai projects are configured to use the same Qdrant collection, chunks from different projects can overwrite each other if they have the same relative file path and chunk index.

The chunker generates chunk IDs from only `filePath` and `chunkIndex`, for example `README.md_0`. The Qdrant backend derives the point UUID from this chunk ID and writes points with Qdrant upsert semantics. As a result, a second project containing `README.md` can overwrite the first project's `README.md` chunk in the shared collection.

This is a security-relevant index integrity issue when a shared Qdrant collection is used for projects that are not mutually trusted. It can cause cross-project vector index poisoning and incorrect search/readback results.

Security note: `SECURITY.md` says security vulnerabilities should not be reported as public issues. This report is formatted using the repository bug template for maintainer submission through the responsible disclosure channel.

### Steps to Reproduce

1. Configure two separate grepai projects to use the same Qdrant collection.
2. In project A, create `README.md` with content:

   ```text
   Victim private deployment runbook
   ```

3. Index project A.
4. In project B, create `README.md` with content:

   ```text
   Attacker decoy repository content
   ```

5. Index project B into the same Qdrant collection.
6. Read back `README.md` from project A's Qdrant-backed store.

E2E verification was performed with the real path:

```text
temporary project file -> Scanner -> Chunker -> Indexer.IndexAll -> QdrantStore.SaveChunks -> real Qdrant container
```

Observed output:

```text
victim_before id="" hash="ab19d91389bc19e2" content="File: README.md\n\nVictim private deployment runbook"
victim_after  id="" hash="5c57e074d9e5a7aa" content="File: README.md\n\nAttacker decoy repository content"
attacker_view id="" hash="5c57e074d9e5a7aa" content="File: README.md\n\nAttacker decoy repository content"

### Expected Behavior

Chunks from different projects should not overwrite each other in a shared vector store. Chunk identity in Qdrant should include a project, repository, workspace, or collection namespace that makes object identity unique across trust domains.

For the example above, project A's `README.md_0` and project B's `README.md_0` should be stored as distinct points or otherwise isolated by collection/project namespace.

### Actual Behavior

Both projects generate the same application-level chunk ID:

```text
README.md_0
```

Qdrant derives the same point UUID from this ID and upserts the point. The second project's chunk overwrites the first project's chunk payload and vector in the shared collection. After project B is indexed, project A reads back project B's content for `README.md`.

Relevant code:

```go
// https://github.com/yoanbernabeu/grepai/blob/65c345ca32122c17a39a5bbec2780c2eea773a12/indexer/chunker.go#L102-L105
hash := sha256.Sum256([]byte(fmt.Sprintf("%s:%d:%d:%s", filePath, pos, end, chunkContent)))
contentHash := sha256.Sum256([]byte(chunkContent))
chunkID := fmt.Sprintf("%s_%d", filePath, chunkIndex)
```

```go
// https://github.com/yoanbernabeu/grepai/blob/65c345ca32122c17a39a5bbec2780c2eea773a12/store/qdrant.go#L117-L145
func (s *QdrantStore) getUUIDForChunk(chunkID string) uuid.UUID {
	namespace := uuid.MustParse("6ba7b810-9dad-11d1-80b4-00c04fd430c8")
	return uuid.NewSHA1(namespace, []byte(chunkID))
}

pointID := s.getUUIDForChunk(chunk.ID)

_, err := s.client.Upsert(ctx, &qdrant.UpsertPoints{
	CollectionName: s.collectionName,
	Points:         points,
})
```

### grepai Version

grepai v0.35.0 commit 65c345ca32122c17a39a5bbec2780c2eea773a12

### Operating System

Linux

### Embedding Provider

Not applicable

### Relevant Logs

```shell
=== RUN   TestSharedQdrantCollectionCollisionE2E
victim_before id="" hash="ab19d91389bc19e2" content="File: README.md\n\nVictim private deployment runbook"
victim_after  id="" hash="5c57e074d9e5a7aa" content="File: README.md\n\nAttacker decoy repository content"
attacker_view id="" hash="5c57e074d9e5a7aa" content="File: README.md\n\nAttacker decoy repository content"
--- PASS: TestSharedQdrantCollectionCollisionE2E (0.34s)
PASS
ok  	grepai-e2e	0.348s
```

### Configuration

```yaml
Minimal affected configuration shape:


store:
  backend: qdrant
  qdrant:
    endpoint: http://localhost
    port: 6334
    collection: shared_collection_name


The issue requires multiple projects to share the same Qdrant collection. Normal single-project default collection naming uses a sanitized project root when `collection` is empty, and workspace indexing prefixes paths/IDs with `workspace/project`, which reduces exposure in those paths.
```

### Checklist

- [ ] I am using the latest version of grepai

3em0:
## Summary
- namespace Qdrant point IDs by project/workspace namespace to avoid same-path chunk overwrites in shared collections
- persist and apply `project_namespace` filters across Qdrant search, scroll, delete, stats, and content-hash lookup operations
- add unit coverage and an optional real-Qdrant integration regression test for shared collection isolation

Closes #247.

## Tests
- `docker run --rm -v /mnt/nas/gitRepo/repos/392:/work -w /work golang:1.25-bookworm go test ./store -v`
- `docker run --rm --network=host -v /mnt/nas/gitRepo/repos/392:/work -w /work -e GREPAI_QDRANT_TEST_ENDPOINT=http://localhost golang:1.25-bookworm go test ./store -run TestQdrantStore_SharedCollectionIsolatedByProjectNamespace -v`
- `docker run --rm --user 1000:1000 -e GOCACHE=/tmp/go-cache -e GOMODCACHE=/tmp/go-mod -v /mnt/nas/gitRepo/repos/392:/work -w /work golang:1.25-bookworm go test ./...`

### DIFF
--- a/cli/search.go
+++ b/cli/search.go
@@ -252,7 +252,7 @@ func runSearch(cmd *cobra.Command, args []string) error {
 			collectionName = store.SanitizeCollectionName(projectRoot)
 		}
 		var err error
-		st, err = store.NewQdrantStore(ctx, cfg.Store.Qdrant.Endpoint, cfg.Store.Qdrant.Port, cfg.Store.Qdrant.UseTLS, collectionName, cfg.Store.Qdrant.APIKey, cfg.Embedder.GetDimensions())
+		st, err = store.NewQdrantStoreWithNamespace(ctx, cfg.Store.Qdrant.Endpoint, cfg.Store.Qdrant.Port, cfg.Store.Qdrant.UseTLS, collectionName, projectRoot, cfg.Store.Qdrant.APIKey, cfg.Embedder.GetDimensions())
 		if err != nil {
 			return fmt.Errorf("failed to connect to qdrant: %w", err)
 		}
@@ -586,7 +586,7 @@ func runWorkspaceSearch(ctx context.Context, query string, projects []string, pa
 		if collectionName == "" {
 			collectionName = "workspace_" + ws.Name
 		}
-		st, err = store.NewQdrantStore(ctx, ws.Store.Qdrant.Endpoint, ws.Store.Qdrant.Port, ws.Store.Qdrant.UseTLS, collectionName, ws.Store.Qdrant.APIKey, ws.Embedder.GetDimensions())
+		st, err = store.NewQdrantStoreWithNamespace(ctx, ws.Store.Qdrant.Endpoint, ws.Store.Qdrant.Port, ws.Store.Qdrant.UseTLS, collectionName, projectID, ws.Store.Qdrant.APIKey, ws.Embedder.GetDimensions())
 		if err != nil {
 			return fmt.Errorf("failed to connect to qdrant: %w", err)
 		}
--- a/cli/status.go
+++ b/cli/status.go
@@ -470,7 +470,7 @@ func runStatus(cmd *cobra.Command, args []string) error {
 			collectionName = store.SanitizeCollectionName(projectRoot)
 		}
 		var err error
-		st, err = store.NewQdrantStore(ctx, cfg.Store.Qdrant.Endpoint, cfg.Store.Qdrant.Port, cfg.Store.Qdrant.UseTLS, collectionName, cfg.Store.Qdrant.APIKey, cfg.Embedder.GetDimensions())
+		st, err = store.NewQdrantStoreWithNamespace(ctx, cfg.Store.Qdrant.Endpoint, cfg.Store.Qdrant.Port, cfg.Store.Qdrant.UseTLS, collectionName, projectRoot, cfg.Store.Qdrant.APIKey, cfg.Embedder.GetDimensions())
 		if err != nil {
 			return fmt.Errorf("failed to connect to qdrant: %w", err)
 		}
--- a/cli/watch.go
+++ b/cli/watch.go
@@ -451,7 +451,7 @@ func initializeStore(ctx context.Context, cfg *config.Config, projectRoot string
 		if collectionName == "" {
 			collectionName = store.SanitizeCollectionName(projectRoot)
 		}
-		return store.NewQdrantStore(ctx, cfg.Store.Qdrant.Endpoint, cfg.Store.Qdrant.Port, cfg.Store.Qdrant.UseTLS, collectionName, cfg.Store.Qdrant.APIKey, cfg.Embedder.GetDimensions())
+		return store.NewQdrantStoreWithNamespace(ctx, cfg.Store.Qdrant.Endpoint, cfg.Store.Qdrant.Port, cfg.Store.Qdrant.UseTLS, collectionName, projectRoot, cfg.Store.Qdrant.APIKey, cfg.Embedder.GetDimensions())
 	default:
 		return nil, fmt.Errorf("unknown storage backend: %s", cfg.Store.Backend)
 	}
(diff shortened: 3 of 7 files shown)


yoanbernabeu:
Thanks for this — the fix is correct and I verified it end to end against a real Qdrant.

## The bug is real and this closes it

Two projects with separate `.grepai/config.yaml` pointing at the same `store.qdrant.collection`, both containing `src/main.go` with different content:

- **On `main`**: project A indexed → 2 points. Project B indexed into the same collection → still **2 points**, and both payloads now hold project B's content. Project A's chunks are gone, silently overwritten.
- **With this PR**: **4 points** coexist, each carrying its `project_namespace`. `grepai search` from A returns only A's content, from B only B's. Isolation confirmed by real searches on both sides.

Worth noting for anyone reading later: this does *not* reproduce in workspace mode, since `projectPrefixStore` already prefixes chunk IDs there. The failure is specific to the single-project mode with a shared collection — exactly what #247 describes.

I also checked the interaction with what landed in `store/qdrant.go` since you opened this: #207 (gRPC max message size at 64 MB, vectors dropped from `GetAllChunks`) and #201 (Windows separators in `sanitizeCollectionName`). Both survive the merge intact, no interference. `go build`, `go vet` and `go test -race ./...` all pass, and the integration test passes against a live Qdrant when `GREPAI_QDRANT_TEST_ENDPOINT` is set.

## What's blocking: existing collections

The PR doesn't handle collections already indexed by the current release, and the result is quiet enough that I don't want to ship it as-is.

Measured: indexed a project with a `main` build (2 points, no `project_namespace` in the payload), then re-ran `grepai watch` with a build of this branch against the **same** collection, source files untouched.

- The watcher re-indexes everything and writes 2 **new** points carrying `project_namespace`.
- The 2 **old** points stay in the collection. Nothing deletes them.
- They are now unreachable forever: `search` filters on `project_namespace` so they never match, and `DeleteByFile` applies the same filter so it can never remove them either.
- `grepai status` / `GetStats` also count through the namespace filter, so the user's stats keep showing the expected number while the actual collection has doubled.

So every existing Qdrant user silently pays double storage after upgrading, with no signal and no way back except dropping the collection and reindexing from scratch. There's no CHANGELOG entry or migration note in the PR either.

The same shape bites on a plain directory rename, since `project_namespace` is the absolute project path — worth considering while you're in there.

## What I'd like before merging

Anything that makes the transition visible would do. In rough order of preference:

1. On startup, detect points missing `project_namespace` in the target collection and log a clear warning telling the user to reindex (a scroll with a "must not have" filter is enough).
2. Better: a one-shot backfill that stamps legacy points with the current project's namespace when the collection has exactly one project's worth of data, and warns otherwise.
3. At minimum: a CHANGELOG entry stating that Qdrant users must drop and reindex their collection after upgrading.

Two small things while you're at it:

- `store/qdrant_integration_test.go` creates `grepai_test_namespace_<timestamp>` collections and never deletes them — each run leaves one behind. A `defer` on delete would keep test instances clean.
- The `Wait: qdrant.PtrOf(true)` added on `Upsert` is a behaviour change (stronger consistency, slightly slower writes) that isn't mentioned in the description. Probably fine, but worth stating so it's a deliberate choice.

Also: your #252 for the Postgres cache landed earlier today. Same class of bug, and the reasoning there carried over well — thanks for both.

Happy to merge this as soon as an upgrading user gets told something.

yoanbernabeu:
Thanks for this — the fix itself is sound, and I verified it against a real Qdrant instance: `TestQdrantStore_SharedCollectionIsolatedByProjectNamespace` passes, `go vet` is clean and the full `go test -race ./...` suite is green.

Before merging I'd like to sort out the upgrade path for people who already have a Qdrant index, because right now the change is silently breaking for them. What I measured, indexing with 0.36.1 and then running the branch against the same collection:

| Step | Result |
| --- | --- |
| Index built with 0.36.1, `search` with 0.36.1 | 2 results |
| Same collection, `search` with this branch | `[]` — the whole index is invisible |
| After a full reindex with this branch | results are back, but the collection went from **2 points to 4** |

Two consequences:

1. **A full reindex becomes mandatory, with no warning.** Point IDs change (`uuid.NewSHA1` now hashes `projectNamespace + "\x00" + chunkID`), so nothing existing resolves any more. For anyone on a paid embedding provider that is a real, unannounced cost.
2. **The old points are never reclaimed.** They carry no `project_namespace` payload, so every filter excludes them — including the delete filters. They stay orphaned in the collection forever and it keeps growing.

Would you be up for adding:

- a startup migration that drops points without a `project_namespace` payload from the collection (or, if you'd rather not delete anything automatically, backfills the payload on them), so upgrading doesn't leave dead weight behind;
- a `CHANGELOG.md` entry under `[Unreleased]` describing the reindex, and a short migration note in the Qdrant docs page.

Happy to take another look as soon as that's in — the isolation logic itself I'm satisfied with.

PR #250, opened 2026-05-20
fix: scope Postgres embedding cache by project

### LINKED ISSUE #249: [Bug]: Cross-project Postgres embedding cache reuse via content_hash-only lookup
### Bug Description

`PostgresStore.LookupByContentHash` performs embedding-cache lookup using only `content_hash`. The `chunks` table stores `project_id` and most Postgres store operations scope reads and writes by `project_id`, but this cache path does not. When multiple projects share one Postgres backend, a project can reuse a vector generated for another project if both chunks have the same raw-content hash.

This is a security-relevant cache key confusion issue in shared Postgres deployments. It can cause cross-project embedding reuse, index integrity loss, search-quality corruption, and possible cache-hit side-channel observations. It does not directly dump the `chunks` table or return source text; the vulnerable query returns only `vector`.

Security note: `SECURITY.md` says security vulnerabilities should not be reported as public issues. This report is formatted using the repository bug template for maintainer submission through the responsible disclosure channel.

### Steps to Reproduce

1. Configure two different grepai projects to use the same Postgres DSN.
2. Index or otherwise store a victim chunk in project `proj_victim` with a non-empty `content_hash`.
3. From a separate project `proj_attacker`, request an embedding cache lookup for the same `content_hash`.
4. Observe that the attacker project receives the victim project's vector because the lookup does not filter by `project_id`.

E2E verification was performed with the official AgentScan PoC using a real `pgvector/pgvector:pg17` Postgres container and the repository's `PostgresStore` code path.

Command:

```shell
cd /home/dem0/桌面/AgentScan/output/audits/17c91c321a7ec16242f3daf50cf500f7/scan_17c91c321a7ec16242f3daf50cf500f7_c07d4db7/pocs/finding-002
docker compose up --abort-on-container-exit --build --exit-code-from tester
docker compose down -v
```

Observed output:

```text
tester-1 | === RUN   TestPoC
tester-1 | VULNERABILITY CONFIRMED: project "proj_attacker" resolved victim vector [9.9 8.8 7.7] via content_hash-only lookup for "sha256:4f9c4a1d2b7e"
tester-1 | --- PASS: TestPoC (0.12s)
tester-1 | PASS
tester-1 | ok   github.com/yoanbernabeu/grepai/store   0.123s
tester-1 exited with code 0
```

The PoC creates two stores over the same Postgres database:

```go
victim, err := NewPostgresStore(ctx, dsn, "proj_victim", 3)
attacker, err := NewPostgresStore(ctx, dsn, "proj_attacker", 3)
```

The victim store writes:

```text
project_id: proj_victim
content_hash: sha256:4f9c4a1d2b7e
vector: [9.9,8.8,7.7]
```

The attacker store then calls `LookupByContentHash(ctx, "sha256:4f9c4a1d2b7e")` and receives `[9.9,8.8,7.7]`.

### Expected Behavior

Embedding cache hits should only be reused when the cached vector is valid for the current security and semantic context. At minimum, Postgres cache lookup should include the current `project_id`. It should also account for embedder identity or cache schema where vectors from different embedding providers, models, dimensions, or embedding input formats are not interchangeable.

For example, project `proj_attacker` should not receive a vector stored under `proj_victim` solely because both chunks share the same raw-content hash.

### Actual Behavior

`LookupByContentHash` queries the shared `chunks` table by `content_hash` only:

```go
// https://github.com/yoanbernabeu/grepai/blob/f6dbf8dbb74c1f80aec37531964665bb71bdc6db/store/postgres.go#L376-L395
func (s *PostgresStore) LookupByContentHash(ctx context.Context, contentHash string) ([]float32, bool, error) {
	if contentHash == "" {
		return nil, false, nil
	}

	var vec pgvector.Vector
	err := s.pool.QueryRow(ctx,
		`SELECT vector FROM chunks WHERE content_hash = $1 AND vector IS NOT NULL LIMIT 1`,
		contentHash,
	).Scan(&vec)
```

The query omits `s.projectID`, even though the same store writes chunks with `project_id` and uses `(project_id, id)` as the conflict target:

```go
// https://github.com/yoanbernabeu/grepai/blob/f6dbf8dbb74c1f80aec37531964665bb71bdc6db/store/postgres.go#L102-L114
`INSERT INTO chunks (id, project_id, file_path, start_line, end_line, content, vector, hash, content_hash, updated_at)
VALUES ($1, $2, $3, $4, $5, $6, $7, $8, $9, $10)
ON CONFLICT (project_id, id) DO UPDATE SET
	file_path = EXCLUDED.file_path,
	start_line = EXCLUDED.start_line,
	end_line = EXCLUDED.end_line,
	content = EXCLUDED.content,
	vector = EXCLUDED.vector,
	hash = EXCLUDED.hash,
	content_hash = EXCLUDED.content_hash,
	updated_at = EXCLUDED.updated_at`,
chunk.ID, s.projectID, chunk.FilePath, chunk.StartLine, chunk.EndLine,
chunk.Content, vec, chunk.Hash, chunk.ContentHash, chunk.UpdatedAt,
```

The indexer trusts the optional `EmbeddingCache` path before calling the embedder:

```go
// https://github.com/yoanbernabeu/grepai/blob/f6dbf8dbb74c1f80aec37531964665bb71bdc6db/indexer/indexer.go#L630-L647
cache, ok := idx.store.(store.EmbeddingCache)
if !ok {
	return nil, 0
}

cached := make(map[int][]float32)
for i, chunk := range chunks {
	if chunk.ContentHash == "" {
		continue
	}
	vec, found, err := cache.LookupByContentHash(ctx, chunk.ContentHash)
```

As a result, a project can reuse a vector generated under a different `project_id`.

### grepai Version

grepai v0.35.0-1-gf6dbf8d commit f6dbf8dbb74c1f80aec37531964665bb71bdc6db

### Operating System

Linux

### Embedding Provider

Not applicable

### Relevant Logs

```shell
=== RUN   TestPoC
VULNERABILITY CONFIRMED: project "proj_attacker" resolved victim vector [9.9 8.8 7.7] via content_hash-only lookup for "sha256:4f9c4a1d2b7e"
--- PASS: TestPoC (0.12s)
PASS
ok   github.com/yoanbernabeu/grepai/store   0.123s
```

### Configuration

```yaml
Minimal affected configuration shape:


store:
  backend: postgres
  postgres:
    dsn: postgres://postgres:postgres@postgres:5432/postgres?sslmode=disable


The issue requires multiple projects or trust domains to share the same Postgres database and `chunks` table. Normal single-user local use has lower security impact. Workspace mode may be less exposed through the current `projectPrefixStore` wrapper because that wrapper does not expose `EmbeddingCache`, but direct multi-project Postgres use is affected.
```

### Checklist

- [ ] I am using the latest version of grepai

3em0:
## Summary
- scope Postgres content-hash embedding cache lookups by `project_id`
- add a project-scoped content-hash index for the new lookup shape
- add SQL and optional Postgres integration regression tests

Fixes #249

## Tests
- `docker run --rm -v /home/dem0/.config/superpowers/worktrees/grepai/fix-postgres-contenthash-cache-249-clean:/repo -w /repo golang:1.24-bookworm go test -mod=mod ./store`
- `docker run --rm --network host -e GREPAI_POSTGRES_TEST_DSN=postgres://postgres:postgres@127.0.0.1:55442/postgres?sslmode=disable -v /home/dem0/.config/superpowers/worktrees/grepai/fix-postgres-contenthash-cache-249-clean:/repo -w /repo golang:1.24-bookworm go test -mod=mod -run TestPostgresStoreLookupByContentHashIsProjectScoped -v ./store`
- `docker run --rm -u $(id -u):$(id -g) -e GOCACHE=/tmp/gocache -e GOMODCACHE=/tmp/gomodcache -v /home/dem0/.config/superpowers/worktrees/grepai/fix-postgres-contenthash-cache-249-clean:/repo -w /repo golang:1.24-bookworm go test -mod=mod ./...`

### DIFF
--- a/store/postgres.go
+++ b/store/postgres.go
@@ -16,6 +16,8 @@ type PostgresStore struct {
 	dimensions int
 }
 
+const lookupByContentHashSQL = `SELECT vector FROM chunks WHERE project_id = $1 AND content_hash = $2 AND vector IS NOT NULL LIMIT 1`
+
 func NewPostgresStore(ctx context.Context, dsn string, projectID string, vectorDimensions int) (*PostgresStore, error) {
 	pool, err := pgxpool.New(ctx, dsn)
 	if err != nil {
@@ -61,7 +63,7 @@ func (s *PostgresStore) ensureSchema(ctx context.Context) error {
 			PRIMARY KEY (project_id, path)
 		)`,
 		`ALTER TABLE chunks ADD COLUMN IF NOT EXISTS content_hash TEXT DEFAULT ''`,
-		`CREATE INDEX IF NOT EXISTS idx_chunks_content_hash ON chunks(content_hash) WHERE content_hash != ''`,
+		`CREATE INDEX IF NOT EXISTS idx_chunks_project_content_hash ON chunks(project_id, content_hash) WHERE content_hash != ''`,
 		buildEnsureVectorSQL(s.dimensions),
 		// Migrate chunks primary key from (id) to (project_id, id) so that
 		// worktrees sharing the same database get their own chunk rows instead
@@ -381,8 +383,8 @@ func (s *PostgresStore) LookupByContentHash(ctx context.Context, contentHash str
 
 	var vec pgvector.Vector
 	err := s.pool.QueryRow(ctx,
-		`SELECT vector FROM chunks WHERE content_hash = $1 AND vector IS NOT NULL LIMIT 1`,
-		contentHash,
+		lookupByContentHashSQL,
+		s.projectID, contentHash,
 	).Scan(&vec)
 
 	if err == pgx.ErrNoRows {
--- a/store/postgres_integration_test.go
+++ b/store/postgres_integration_test.go
@@ -1,0 +1,70 @@
+package store
+
+import (
+	"context"
+	"os"
+	"testing"
+	"time"
+)
+
+func TestPostgresStoreLookupByContentHashIsProjectScoped(t *testing.T) {
+	dsn := os.Getenv("GREPAI_POSTGRES_TEST_DSN")
+	if dsn == "" {
+		t.Skip("GREPAI_POSTGRES_TEST_DSN is not set")
+	}
+
+	ctx := context.Background()
+	victim, err := NewPostgresStore(ctx, dsn, "proj_victim", 3)
+	if err != nil {
+		t.Fatalf("failed to create victim store: %v", err)
+	}
+	defer victim.Close()
+
+	attacker, err := NewPostgresStore(ctx, dsn, "proj_attacker", 3)
+	if err != nil {
+		t.Fatalf("failed to create attacker store: %v", err)
+	}
+	defer attacker.Close()
+
+	if _, err := victim.pool.Exec(ctx, `TRUNCATE TABLE chunks, documents`); err != nil {
+		t.Fatalf("failed to clean tables: %v", err)
+	}
+
+	contentHash := "sha256:shared-content"
+	victimVector := []float32{9.9, 8.8, 7.7}
+	if err := victim.SaveChunks(ctx, []Chunk{{
+		ID:          "README.md_0",
+		FilePath:    "README.md",
+		StartLine:   1,
+		EndLine:     1,
+		Content:     "File: README.md\n\nShared content",
+		Vector:      victimVector,
+		Hash:        "victim-row",
+		ContentHash: contentHash,
+		UpdatedAt:   time.Now().UTC(),
+	}}); err != nil {
+		t.Fatalf("failed to save victim chunk: %v", err)
+	}
+
+	if _, found, err := attacker.LookupByContentHash(ctx, contentHash); err != nil {
+		t.Fatalf("attacker lookup failed: %v", err)
+	} else if found {
+		t.Fatal("attacker project should not reuse victim project's cached vector")
+	}
+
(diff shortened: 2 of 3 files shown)


yoanbernabeu:
Thank you for this fix, and for the thorough reproduction in #249 — the diagnosis was correct and the project-scoped index you added was the right shape.

I'm closing this in favour of #252, which covers the same ground plus the second half of the problem (#251): the cache also needs to be scoped by *embedder identity*, not only by project.

I verified the gap against a real PostgreSQL instance. With this PR's code, two stores sharing the same `project_id` but representing two different embedder configurations (different provider/model, same vector dimensions) still hit each other's cache:

```
storeB (different embedder config, same project) reused storeA's vector: [1.1 2.2 3.3]
```

`project_id` alone can't catch that, since both rows legitimately belong to the same project. #252 adds an `embedding_cache_namespace` derived from provider/model/dimensions/endpoint, which closes both #249 and #251 in one lookup.

I also checked that #252 is safe for existing deployments: on a database populated by the current `main`, the upgrade applied `ALTER TABLE chunks ADD COLUMN IF NOT EXISTS embedding_cache_namespace` and created the new index cleanly, with **no reindexing** (`0 files indexed, 4 skipped`) and identical search results.

This is not a reflection on your patch — it was correct and well-tested for the issue it targeted. Two contributors landed overlapping fixes and I took the superset. Thanks for the time you put into it, and please do send more.

PR #258, opened 2026-05-28
feat(scanner): index projects rooted at a symlink

kryptt:
## Problem

\`filepath.WalkDir\` does not descend into symlinked directories, so when a user points grepai at a project root that is itself a symlink, the scan silently enumerates zero files. There's no warning — just no results.

This is hostile to a common pattern: fronting a version-managed path with a stable name so the workspace survives upstream version bumps. For example:

\`\`\`
~/.emacs.d/emacs-sources -> /usr/share/emacs/30.2/
\`\`\`

A user puts \`emacs-sources\` in their workspace config so they don't have to retarget on every emacs upgrade. Today, that workspace indexes nothing.

## Solution

\`NewScanner\` now \`os.Lstat\`s the root and, only if it is a symlink to a directory, resolves it once via \`filepath.EvalSymlinks\` before \`WalkDir\` runs. Intermediate symlinks deeper in the tree are still skipped — that matches upstream's existing behavior and avoids recursion loops or surprising blow-ups when packages vendor third-party trees via symlinks.

We deliberately do **not** call \`EvalSymlinks\` on a real directory: doing so would also rewrite parent-directory symlinks (e.g. \`/usr/local\` on a Nix system) and silently shift the project's stored path in ways the caller never asked for. The \`os.Lstat\` guard makes the resolution opt-in based on the root entry's own type.

## Test plan

Two new tests in \`indexer/scanner_test.go\`:

| Test | Asserts |
|---|---|
| \`TestScanner_SymlinkProjectRoot\` | a symlink-to-directory root enumerates the underlying files; relative paths returned are relative to the resolved real root (not the symlink path) |
| \`TestScanner_RealDirRoot_NotEvalSymlinks\` | when the root is a real directory (even if its parent is reached through a symlink), \`scanner.root\` stays as the caller-provided path, no silent rewrite |

Both tests \`t.Skipf\` cleanly on environments that cannot create symlinks. All existing tests continue to pass (\`go test ./...\`).

### DIFF
--- a/indexer/scanner.go
+++ b/indexer/scanner.go
@@ -117,6 +117,20 @@ type Scanner struct {
 }
 
 func NewScanner(root string, ignore *IgnoreMatcher) *Scanner {
+	// Resolve the project root if (and only if) it is itself a symlink to a
+	// directory. filepath.WalkDir does not descend into symlinked dirs, so a
+	// project rooted at a symlink would otherwise enumerate zero files. We
+	// only resolve the root entry — intermediate symlinks deeper in the tree
+	// remain skipped, which avoids recursion loops and surprising blow-ups
+	// when packages vendor third-party trees through symlinks. We deliberately
+	// do not call filepath.EvalSymlinks unconditionally because that would
+	// also rewrite parent-directory symlinks (e.g. /usr/local on a Nix system)
+	// and shift the project's stored path in ways the user didn't ask for.
+	if info, err := os.Lstat(root); err == nil && info.Mode()&os.ModeSymlink != 0 {
+		if resolved, err := filepath.EvalSymlinks(root); err == nil {
+			root = resolved
+		}
+	}
 	return &Scanner{
 		root:   root,
 		ignore: ignore,
--- a/indexer/scanner_test.go
+++ b/indexer/scanner_test.go
@@ -352,3 +352,79 @@ func TestScanner_ScanFile_SkipsMinified(t *testing.T) {
 		t.Error("expected nil for minified file, got file info")
 	}
 }
+
+func TestScanner_SymlinkProjectRoot(t *testing.T) {
+	// Real directory containing source files.
+	realDir := t.TempDir()
+	if err := os.WriteFile(filepath.Join(realDir, "main.go"), []byte("package main"), 0644); err != nil {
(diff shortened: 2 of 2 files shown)


yoanbernabeu:
Thanks for this — the change is small and well argued, and I like that it resolves the root entry only rather than calling `EvalSymlinks` unconditionally.

Before merging I tried to reproduce the failure it fixes, and I couldn't. What I did, on macOS:

```
mkdir -p /tmp/realdir/src && echo 'package src ...' > /tmp/realdir/src/api.go
ln -s /tmp/realdir /tmp/linkroot
cd /tmp/linkroot && grepai init --yes -p ollama -b gob && grepai watch --background
```

Both `main` **and** the 0.36.1 binary from before today's merges report `Files indexed: 1` and `grepai search` returns `src/api.go`. So in this shape the symlinked root enumerates fine without your patch.

My guess is that my scenario isn't yours: entering the symlink with `cd` means Go's `os.Getwd()` hands the already-resolved path to the scanner, so `filepath.WalkDir` never sees the symlink. Presumably you hit it where the root arrives from somewhere other than the working directory — a workspace entry pointing at a symlink, a path passed explicitly, or a specific platform. Could you share the exact repro? I'd like to see the failure before merging the fix, and to add a regression test that actually covers your path.

Related, and possibly relevant to how you want this to land: #300 (merged today) added a per-project writer lock whose `canonicalProjectRoot` calls `filepath.EvalSymlinks` on the project root **unconditionally**, and `watchProjectWithEventObserver` now passes that resolved path down to the rest of the watch path. That is the broader resolution your comment explicitly argues against — it rewrites parent-directory symlinks too, the `/usr/local`-on-Nix case you mention. So depending on your repro, this PR may now be either redundant for the watch path, or the right place to reconsider how aggressively we canonicalize. Your read on that would be useful.

Merge conflict status: the only conflict with current `main` is in `indexer/scanner_test.go`, and it is purely additive — your `TestScanner_SymlinkProjectRoot` and `TestScanner_RealDirRoot_NotEvalSymlinks` against the custom-extension tests that landed since. I resolved it locally keeping both and the full suite is green (1480 tests), so that part is not a blocker.

PR #259, opened 2026-05-28
feat(workspace): relocate per-project local state to ~/.grepai/

kryptt:
## Problem

When a workspace indexes a read-only project root — typical examples are \`/usr/share/emacs/<X.Y>/\`, \`/opt/foo/\`, or any directory owned by root and read-only for the user — grepai's attempt to \`mkdir <project_root>/.grepai/\` fails with \`EACCES\`. The symbol index and RPG index silently fail to persist; the watcher logs a warning per project on every run.

In my own workspace this affected \`emacs-source\` (\`/usr/share/emacs/30.2\`) and \`site-lisp\` (\`/usr/share/emacs/site-lisp\`). The PostgreSQL vector store still landed chunks fine, but the per-project local state was unreachable.

## Solution

Adds a workspace-level \`local_state_dir\` option:

\`\`\`yaml
workspaces:
  emacs:
    local_state_dir: ~/grepai-state   # or any absolute path
    projects:
      - name: emacs-source
        path: /usr/share/emacs/30.2
\`\`\`

- When set: each project's local state lives at \`<local_state_dir>/<project-name>/\`. A leading \`~/\` is expanded against the user's home.
- When **unset** (the new default): local state lives under \`~/.grepai/workspaces/<workspace-name>/<project-name>/\`. Workspace projects no longer touch the project root at all.

Two helpers — \`Workspace.ResolvedLocalStateDir()\` and \`Workspace.ProjectStateDir(projectName)\` — encapsulate the expansion + join logic and are useful for downstream consumers (e.g. future MCP tooling that wants to read the symbol index).

The workspace runner in \`cli/watch.go\` \`MkdirAll\`s the project state dir and feeds it to the GOB symbol + RPG stores instead of \`config.GetSymbolIndexPath(project.Path)\` / \`config.GetRPGIndexPath(project.Path)\`. Those helpers are unchanged and still target \`<project_root>/.grepai/\` for the single-project flow where it makes sense.

Single-project mode (\`grepai watch\` without \`--workspace\`) and per-project \`.grepai/config.yaml\` are unaffected — only the workspace runner changes location.

## Behavior change

Before: workspace projects created (or attempted to create) \`<project_root>/.grepai/\`.
After: workspace projects use \`~/.grepai/workspaces/<workspace-name>/<project-name>/\` by default.

Existing workspace users will see their per-project state move to a new location on first run after upgrade. Vector stores are unaffected (those go to PostgreSQL/Qdrant); only the GOB symbol/RPG indexes relocate. Each will rebuild on the next \`grepai watch --workspace …\`.

If preserving in-place behavior is preferred, the field can default to empty-with-fallback-to-project-root instead; happy to adjust based on review.

## Test plan

Four new tests in \`config/workspace_test.go\`:

| Test | Asserts |
|---|---|
| \`TestWorkspace_ResolvedLocalStateDir_Default\` | unset → \`~/.grepai/workspaces/<name>/\` |
| \`TestWorkspace_ResolvedLocalStateDir_TildeExpansion\` | \`~/foo\` expands to \`\$HOME/foo\` |
| \`TestWorkspace_ResolvedLocalStateDir_AbsolutePassthrough\` | \`/var/lib/...\` passes through verbatim |
| \`TestWorkspace_ProjectStateDir\` | composes resolved base with project name |

All existing tests continue to pass (\`go test ./...\`).

### DIFF
--- a/cli/watch.go
+++ b/cli/watch.go
@@ -2763,7 +2763,15 @@ func initializeWorkspaceRuntime(ctx context.Context, ws *config.Workspace, proje
 	}
 	idx := indexer.NewIndexer(project.Path, vectorStore, emb, chunker, scanner, projectCfg.Watch.LastIndexTime, processorRegistry)
 	extractor := trace.NewRegexExtractor()
-	symbolStore := trace.NewGOBSymbolStore(config.GetSymbolIndexPath(project.Path))
+	// Workspace projects keep their local state (symbol index, RPG index)
+	// under the workspace-scoped state dir rather than inside the project
+	// root. This lets us index read-only trees like /usr/share/emacs/30.2/
+	// without needing write access to the project root.
+	projectStateDir := ws.ProjectStateDir(project.Name)
+	if err := os.MkdirAll(projectStateDir, 0755); err != nil {
+		log.Printf("Warning: failed to create state dir %s for %s: %v", projectStateDir, project.Name, err)
+	}
+	symbolStore := trace.NewGOBSymbolStore(filepath.Join(projectStateDir, config.SymbolIndexFileName))
 	if err := symbolStore.Load(ctx); err != nil {
 		log.Printf("Warning: failed to load symbol index for %s: %v", project.Path, err)
 	}
@@ -2789,7 +2797,7 @@ func initializeWorkspaceRuntime(ctx context.Context, ws *config.Workspace, proje
 	var rpgEncoder *rpg.RPGEncoder
 	var manager *rpgRealtimeManager
 	if projectCfg.RPG.Enabled {
-		rpgStore = rpg.NewGOBRPGStore(config.GetRPGIndexPath(project.Path))
+		rpgStore = rpg.NewGOBRPGStore(filepath.Join(projectStateDir, config.RPGIndexFileName))
 		if err := rpgStore.Load(ctx); err != nil {
 			log.Printf("Warning: failed to load RPG index for %s: %v", project.Path, err)
 		}
--- a/config/workspace.go
+++ b/config/workspace.go
@@ -25,7 +25,43 @@ type Workspace struct {
 	Name     string         `yaml:"name"`
 	Store    StoreConfig    `yaml:"store"`
 	Embedder EmbedderConfig `yaml:"embedder"`
-	Projects []ProjectEntry `yaml:"projects"`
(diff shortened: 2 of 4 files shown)


yoanbernabeu:
Thanks for this PR — the read-only project root problem is real (`/usr/share/emacs/<X.Y>/`, `/opt/...`), and a workspace-level `local_state_dir` is the right shape for it.

One blocker before this can go in: **the default is not backward compatible**, and it affects every workspace user, not just the ones with a read-only root.

## What I measured

With a workspace on Qdrant and an ordinary (writable, non-symlink) project root:

- Binary built from `main`: symbol index written to `<project_root>/.grepai/symbols.gob` (2 symbols extracted)
- Binary built from this branch, **same workspace, no config change at all**: a new `symbols.gob` is created under `~/.grepai/workspaces/<ws>/<project>/` and the 2 symbols are re-extracted from scratch. The old `<project_root>/.grepai/symbols.gob` stays on disk, orphaned — never migrated, never cleaned up, and no log line tells the user anything moved.

Because `local_state_dir` is empty by default, `ResolvedLocalStateDir()` falls through to `~/.grepai/workspaces/...` for everyone, with no opt-in.

## Impact on existing users

- The vector store (Postgres/Qdrant) is untouched, so `grepai search` keeps working — good.
- The symbol index and the RPG index are silently invalidated and rebuilt on the next `grepai watch --workspace`. On a large workspace that is a real CPU/time cost, paid without warning.
- Stale `symbols.gob` / `rpg.gob` files are left behind in every project root of the workspace.

## What I'd like instead

You already proposed the fix yourself in the PR description, and it is the one I'd go with: **when `local_state_dir` is empty, keep the historical location** `<project_root>/.grepai/`, and only relocate when the user explicitly sets the key. That way:

- existing workspaces are bit-for-bit unaffected — no rebuild, no orphaned files
- the read-only-root use case is served by opting in, which is exactly the situation where the user already knows they need it

If you'd rather keep `~/.grepai/` as the eventual default, that's a conversation worth having on its own, but it needs a migration path (move the existing files, or at least detect and log them) rather than a silent switch.

Everything else looks good: `go build`, `go vet` and `go test -race ./...` all pass, the config plumbing is clean, and loading a pre-existing `workspace.yaml` without the new key works fine.

Happy to merge once the empty value preserves the current behaviour.

PR #260, opened 2026-05-29
feat(trace): wire up --mode dispatch and ship tree-sitter by default

kryptt:
## Problem

There are two related bugs in current `main` around the `treesitter` integration:

1. **`--mode` does nothing.** `cli/trace.go` exposes `--mode fast` / `--mode precise`. The value is stored on `TraceResult.Mode` and surfaces in JSON output, but no code consults it to pick an extractor. Both `cli/watch.go:1006` and `cli/watch.go:2765` hardcode `trace.NewRegexExtractor()`; `cli/trace.go` reads the symbol store built by watch and never instantiates an extractor. The user-visible result: passing `--mode precise` produces the same symbols `--mode fast` would.

2. **Default users never get tree-sitter.** Every `trace/extractor_ts.go`, every `trace/extractor_ts_*_test.go`, and the only call site through `fsharp/` is gated behind `//go:build treesitter`. Stock `go build` and `make build` (no flags, no `BUILD_TAGS`) produce a binary that lacks the entire tree-sitter path. F# (PR #152) and the rest of the precise-mode work landed but are dead code in any binary not built with the tag explicitly.

This PR fixes both at once.

### 1. Tree-sitter is always compiled in

Drop `//go:build treesitter` from `trace/extractor_ts.go` and its three test files. The grammars become unconditional dependencies via `github.com/smacker/go-tree-sitter` (already in `go.sum`) and the vendored `fsharp/` package. Binary grows from ~23 MB to ~48 MB on x86_64-linux. Comparable to other code-search tools (`ast-grep`, `rg`).

### 2. New `CompoundExtractor` dispatches per-mode

\`trace/compound.go\` adds `CompoundExtractor`, which implements the existing `SymbolExtractor` interface by routing each file between `TreeSitterExtractor` and `RegexExtractor`:

\`\`\`go
case ModeFast:    // -> always RegexExtractor
case ModePrecise: // -> TreeSitterExtractor; error on unsupported extension
default:          // ModeAuto: TS if grammar exists, else regex
\`\`\`

\`trace/languages.go\` exposes `HasTreeSitterGrammar(ext) bool` and `TreeSitterExtensions() []string` as the single source of truth for which extensions are tree-sitter-backed in this build. Adding a new tree-sitter grammar in a future PR is one entry in that map plus the parser registration in `NewTreeSitterExtractor` — the dispatch picks it up automatically.

### 3. `--mode auto` is the new default

CLI flag default in `cli/trace.go` flips from `\"fast\"` to `\"auto\"`. The same flip happens in `config.DefaultConfig().Trace.Mode`. The help text now describes all three modes. \`trace.ParseMode\` normalizes empty/whitespace input to \`auto\` and emits a one-line stderr warning for unknown values.

### 4. Watch flow uses the compound extractor

Both `runWatch` (single-project) and the workspace-mode runner now call `trace.NewCompoundExtractor(trace.ParseMode(cfg.Trace.Mode))` instead of `trace.NewRegexExtractor()`. Function signatures that took \`*trace.RegexExtractor\` widen to \`trace.SymbolExtractor\` — the interface that already exists. No new dependency injection plumbing.

## What the user sees

- The same workspace, after upgrade, gets tree-sitter-quality symbols for `.go`, `.js`, `.jsx`, `.ts`, `.tsx`, `.py`, `.php`, `.cs`, `.fs`, `.fsx`, `.fsi` without changing config. For everything else, regex behavior is unchanged.
- \`grepai trace callers FOO\` and friends finally respect \`--mode precise\`.
- \`--mode unknownvalue\` warns once on stderr and continues with \`auto\` instead of failing.

## Test plan

- New \`trace/compound_test.go\` covers the dispatch matrix:
  - \`ParseMode\` normalization (empty / case / whitespace / unknown-falls-back)
  - \`ModeAuto\`: \`.go\` → TS, \`.lua\` → regex, \`.xyz\` → regex
  - \`ModeFast\`: every file → regex
  - \`ModePrecise\`: \`.go\` → TS, \`.lua\` → error with \`--mode precise\` and \`.lua\` in the message
  - End-to-end `ExtractSymbols` on a sample `.go` file producing the expected named symbols
  - `Mode()` returns the configured string for `TraceResult.Mode` display
  - `SupportedLanguages()` returns the sorted union of both underlying extractors

- The existing \`trace/extractor_ts_csharp_test.go\`, \`trace/extractor_ts_fsharp_test.go\`, and \`trace/extractor_ts_docstring_test.go\` (previously build-tag-gated) now run unconditionally and continue to pass.

- \`go test ./...\` is clean.

## Things to think about during review

- **Binary size growth.** ~23 MB → ~48 MB on x86_64-linux. If this is a problem we could add a `make build-minimal` target that re-introduces the build tag in the opposite direction (default = full; opt-out = small). Happy to do that in a follow-up if you want.
- **Cross-compilation.** CGo is now in the default build path. The F# PR (#152) already required this; CI presumably handles it. Worth confirming the release matrix isn't surprised.
- **No new languages here on purpose.** This PR is intentionally restricted to plumbing so review can focus on dispatch correctness. A follow-up PR will use this foundation to add tree-sitter coverage for the union of (smacker-shipped grammars) ∩ (extensions already in `indexer/scanner.SupportedExtensions`), plus a vendored Wilfred `tree-sitter-elisp` for `.el`.

Closes nothing directly, but unblocks the latent intent of #152 and #176 (Lua fast-mode) by making the dispatch actually live.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -9,6 +9,14 @@ on:
 permissions:
   contents: read
 
+# The tree-sitter grammars are C. Native runners (ubuntu, macos, windows)
+# all ship a C toolchain, so `go build`/`go test` work as-is — but be
+# explicit rather than rely on the default, since a CGO-disabled build
+# fails with a confusing "build constraints exclude all Go files" instead
+# of a missing-compiler error.
+env:
+  CGO_ENABLED: "1"
+
 jobs:
   lint:
     name: Lint
@@ -72,13 +80,14 @@ jobs:
           go-version: "1.25"
           cache: true
 
+      # Native build only. The tree-sitter grammars are C, so every target
+      # is a CGO build and cross-compiling one needs that target's C
+      # toolchain — which the bare runner does not have. The Test job above
+      # already builds and runs the full suite natively on ubuntu, macos and
+      # windows; the cross-compiled release binaries are produced by
+      # release.yml inside the goreleaser-cross image.
       - name: Build
-        run: |
-          GOOS=linux GOARCH=amd64 go build -o grepai-linux-amd64 ./cmd/grepai
-          GOOS=linux GOARCH=arm64 go build -o grepai-linux-arm64 ./cmd/grepai
-          GOOS=darwin GOARCH=amd64 go build -o grepai-darwin-amd64 ./cmd/grepai
-          GOOS=darwin GOARCH=arm64 go build -o grepai-darwin-arm64 ./cmd/grepai
-          GOOS=windows GOARCH=amd64 go build -o grepai-windows-amd64.exe ./cmd/grepai
+        run: go build -o grepai-linux-amd64 ./cmd/grepai
 
       - name: Upload artifacts
         uses: actions/upload-artifact@v7
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -17,18 +17,25 @@ jobs:
         with:
           fetch-depth: 0
 
-      - name: Set up Go
-        uses: actions/setup-go@v7
-        with:
-          go-version-file: go.mod
-          cache: true
-
-      - name: Run GoReleaser
-        uses: goreleaser/goreleaser-action@v7
-        with:
-          distribution: goreleaser
-          version: latest
-          args: release --clean
+      # The tree-sitter grammars are C, so every release target is a CGO
+      # build and needs its own C toolchain. goreleaser-action runs on the
+      # bare runner, which only has the host toolchain, so GoReleaser runs
+      # inside goreleaser-cross instead: it bundles osxcross for darwin,
+      # aarch64 cross-gcc for linux/arm64 and mingw-w64 for windows. The
+      # image tag pins both the Go version (must track go.mod) and the
+      # GoReleaser version.
+      #
+      # Note: windows/arm64 is no longer produced. goreleaser-cross ships no
+      # arm64 mingw toolchain, so that target cannot be a CGO build.
+      - name: Run GoReleaser (CGO cross)
+        run: |
+          docker run --rm \
+            -v "$PWD":/src \
+            -w /src \
+            -e GITHUB_TOKEN \
+            -e HOMEBREW_TAP_GITHUB_TOKEN \
+            ghcr.io/goreleaser/goreleaser-cross:v1.25.v2.12.7 \
+            release --clean
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
           HOMEBREW_TAP_GITHUB_TOKEN: ${{ secrets.GH_PAT }}
--- a/.goreleaser.yml
+++ b/.goreleaser.yml
@@ -4,18 +4,70 @@ before:
   hooks:
     - go mod download
 
+# The tree-sitter grammars are C, so releases build with CGO_ENABLED=1 and
+# every target needs its own cross toolchain. Those toolchains come from the
+# ghcr.io/goreleaser/goreleaser-cross image that release.yml runs this config
+# inside — osxcross for darwin, aarch64 cross-gcc for linux/arm64, mingw-w64
+# for windows. One build entry per target keeps each CC/CXX pairing visible
+# rather than hidden in a nested template.
 builds:
-  - main: ./cmd/grepai
+  - id: linux-amd64
+    main: ./cmd/grepai
     binary: grepai
+    goos: [linux]
+    goarch: [amd64]
     env:
-      - CGO_ENABLED=0
-    goos:
-      - linux
-      - darwin
-      - windows
-    goarch:
-      - amd64
-      - arm64
+      - CGO_ENABLED=1
+      - CC=x86_64-linux-gnu-gcc
+      - CXX=x86_64-linux-gnu-g++
+    ldflags:
+      - -s -w -X main.version={{.Version}}
+
+  - id: linux-arm64
+    main: ./cmd/grepai
+    binary: grepai
+    goos: [linux]
+    goarch: [arm64]
+    env:
+      - CGO_ENABLED=1
+      - CC=aarch64-linux-gnu-gcc
+      - CXX=aarch64-linux-gnu-g++
+    ldflags:
+      - -s -w -X main.version={{.Version}}
+
+  - id: darwin-amd64
+    main: ./cmd/grepai
+    binary: grepai
+    goos: [darwin]
+    goarch: [amd64]
+    env:
+      - CGO_ENABLED=1
+      - CC=o64-clang
+      - CXX=o64-clang++
+    ldflags:
+      - -s -w -X main.version={{.Version}}
+
+  - id: darwin-arm64
+    main: ./cmd/grepai
+    binary: grepai
+    goos: [darwin]
+    goarch: [arm64]
+    env:
+      - CGO_ENABLED=1
+      - CC=oa64-clang
+      - CXX=oa64-clang++
+    ldflags:
+      - -s -w -X main.version={{.Version}}
+
+  - id: windows-amd64
+    main: ./cmd/grepai
+    binary: grepai
+    goos: [windows]
+    goarch: [amd64]
+    env:
+      - CGO_ENABLED=1
+      - CC=x86_64-w64-mingw32-gcc
+      - CXX=x86_64-w64-mingw32-g++
     ldflags:
       - -s -w -X main.version={{.Version}}
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,21 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Changed
+
+- **Tree-sitter symbol extraction is now always compiled in** — the `treesitter` build tag is gone. Stock `go build`/`make build` ships with grammars for the existing 9 languages (Go, JS/JSX, TS/TSX, Python, PHP, C#, F#). Binary size grows from ~23 MB to ~48 MB.
+- **`--mode` flag is now wired up.** Previously `--mode fast`/`--mode precise` was a cosmetic label that never actually selected an extractor; both code paths ran the regex extractor. Three values are now meaningful:
+  - `auto` (new default) — tree-sitter for languages with a compiled-in grammar; regex everywhere else
+  - `fast` — force every file through the regex extractor
+  - `precise` — force every file through the tree-sitter extractor; errors on extensions without a grammar (testing aid)
+- **`grepai watch` (single-project and workspace modes) now uses the compound extractor by default.** Symbol indexes built by the watcher gain whatever tree-sitter sees that regex missed. No user action required.
+
+### Added
+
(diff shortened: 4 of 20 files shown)


yoanbernabeu:
Thanks for this series — the infrastructure work is genuinely good. The `LangSpec`/`NamedQuery` registry is a clean way to add grammars, the vendored `elisp/` code is properly licensed (MIT, Wilfred/tree-sitter-elisp) with provenance recorded, and I confirmed the stack is linear: #262 contains #261 contains #260, same commit SHAs, so merging #262 is exactly merging all three.

I tested the series and found two problems that go to the heart of what it advertises. Posting here since #260 is the base; #261 and #262 inherit both.

## 1. `--mode` is still cosmetic for `grepai trace`

`trace.NewCompoundExtractor` is called from exactly one place in the tree: `cli/watch.go:2014`. **`cli/trace.go` never calls it.** `runTraceCallers` reads the already-persisted `GOBSymbolStore`, and the `--mode` flag it accepts only ends up as a `"mode": "..."` string in the JSON output.

Reproduced: set `mode: fast` in config, rebuilt the index with `watch`, then ran `trace callers validate --json --mode precise` → byte-identical to `--mode fast`, including the same regex-typical self-call false positive. Only the mode configured at the last `watch` has any effect.

So the headline claim — "`--mode` finally dispatches" — holds for the watch-side index build (that part is real and works), but not for the `trace` commands themselves.

## 2. Most of the new languages are inert, and call graphs are broken for the rest

**`config.DefaultConfig().Trace.EnabledLanguages` was never updated.** It still reads `.go .js .ts .jsx .tsx .vue .py .php .lua .c .h .cpp .hpp .cc .cxx .hxx .rs .zig .cs .java .fs .fsx .fsi .pas .dpr`. Of the 16 languages added in #261, only C, C++, Rust, Java and Lua are reachable by default; **Ruby, Scala, Bash, Kotlin, Swift, SQL, Protobuf, HCL, Elm, TOML and Emacs Lisp — 11 of 16 — are compiled into the binary but never extracted**, unless the user hand-edits `enabled_languages` (which the PR doesn't document).

Reproduced: a `.rb` file with two methods, indexed, then `trace callers format_message --json` → `{"query": "format_message", "mode": "auto"}`, no symbol.

**And for the languages that *are* enabled, `trace callers`/`callees` is broken or degraded.** `walkNodeForReferences` only recognises `call_expression`/`invocation_expression`, and `findContainingFunction` only `function_declaration`/`method_declaration`/`constructor_declaration`/`function_definition`/`local_function_statement` — node names inherited from the original 9 languages.

- **Java** (enabled by default, listed as top-priority in #262): a call node is `method_invocation`, which isn't in the list. `trace callers helperMethod` returns **zero callers** even though `run()` calls it in your own fixture.
- **Rust**: `call_expression` matches by coincidence, but the function node is `function_item`, which isn't recognised — the caller comes back as a `<top-level>` placeholder instead of `main`.

The reason CI is green is that `trace/extractor_lang_test.go` only exercises `ExtractSymbols`, never `ExtractReferences`. Symbol extraction is well covered; the call graph — the feature `trace` exists for — isn't tested at all for the new languages.

## Cost side

Binary goes from 33 MB to 71 MB (×2.1), and build time roughly doubles because of cgo. That's a fair price for working multi-language call graphs; it's a hard sell for grammars that are compiled in but never reached.

Also worth noting: existing users won't see any of this until they delete `.grepai/symbols.gob`, since `runInitialScan` skips files unchanged since `lastIndexTime` with no notion of extractor version. (Related: #264 tried to address exactly this and turned out to be a no-op for the same reason.)

One small inconsistency: the PR body pins elisp at `3a8f590` while `elisp/README.md` says `29b4e49275f4a947ce17c8533bc20a1f97768c70`. Not blocking, but worth reconciling.

## What I'd need to merge this

1. Have `cli/trace.go` actually instantiate the configured extractor, so `--mode` means something there — or drop the flag and document that mode is a watch-time setting.
2. Add the new extensions to `EnabledLanguages` (or explain in the docs how to enable them).
3. Extend `walkNodeForReferences` / `findContainingFunction` with the node kinds for the new languages — at minimum Java's `method_invocation` and Rust's `function_item`, since those two ship enabled.
4. Add `ExtractReferences` tests for the new languages, so a broken call graph fails CI instead of passing it.

None of this is wasted work — the registry, the grammars and the licensing are the hard part and they're done. It's the wiring at the two ends that needs closing before the feature does what the PRs say it does. Happy to re-review once that's in.

kryptt:
Thank you — both findings were correct, and the second one was worse than you described. All four items are addressed; the stack is rebased on current `main` and all three PRs are mergeable again.

## 1. `--mode` now dispatches for `grepai trace`

You were right that `cli/trace.go` never instantiated an extractor. Digging into why, the honest answer turned out to be structural: `trace` answers from the index `watch` built, and an index built in `fast` mode contains no tree-sitter symbols at all — there is nothing for a precise query to find. So making the flag mean something meant re-extracting.

`GOBSymbolStore` now records the mode that produced its symbols (persisted; an older index decodes as `""` and is treated as *unknown*, not as any particular mode). `cli/trace_mode.go` re-extracts the project with the requested extractor into a throwaway store when — and only when — `--mode` is passed explicitly *and* differs from the index's recorded mode. Omitting `--mode` costs nothing, which keeps the common path at its current speed. Re-extraction prints a note to stderr since the cost scales with the repo:

```console
$ grepai trace callers validate --mode precise
Note: index was built in "fast" mode; re-extracting /path/to/repo in "precise" mode
      (set trace.mode and re-run `grepai watch` to make this permanent)
```

And `TraceResult.Mode` now names the extractor the answer actually came from rather than echoing the flag, so the JSON field stops lying either way. Workspace-scoped trace gets the same treatment per project.

If you would rather not carry the re-extraction path at all, dropping the flag and documenting mode as watch-time-only is a smaller diff and I am happy to switch — say the word.

## 2. `EnabledLanguages`

Fixed, and it was worse than the count you gave: the list also predated `.mjs`/`.cjs`/`.mts`/`.cts` and `.mill`, which had no parser route either. There is now a single `config.DefaultTracedLanguages` covering every compiled-in grammar plus the regex-only extras, replacing the three hand-maintained copies (`DefaultConfig` and two `cli/watch.go` fallbacks).

## 3. Node kinds — and two of the original nine were broken too

Rather than grow those two switches, the node types moved into `LangSpec`: `CallNodes`, `CalleeFields`, `FunctionNodes`, `FunctionNameFields`. A grammar now declares its own node kinds next to its queries, and `walkNodeForReferences` / `findContainingFunction` read the registry. Every entry was verified by dumping the actual parse tree for a sample file rather than inferred — which is how I found the following:

- **Java**: `method_invocation` (field `name`), plus `method_declaration`/`constructor_declaration`. Your `helperMethod` repro now returns both callers.
- **Rust**: `function_item`, so the caller is `main` rather than `<top-level>`.
- **Python**: the call node is `call`, **not** `call_expression`. Python has never produced a single call reference. Fixed.
- **PHP**: `function_call_expression` / `member_call_expression`. Same story. Fixed.
- **JS/TS**: gained `method_definition`, so class methods stop reporting `<top-level>`.
- C and C++ nest the function name under `declarator`; Lua's callee is several sibling `prefix` children (`t.helper(1)` is three); Kotlin's and Swift's `call_expression` carry no fields at all. Each is a line of registry data with a comment saying why.

Six languages declare **no** `CallNodes` on purpose: SQL, Protobuf, HCL, Elm and TOML are declarative, and in Emacs Lisp a call is an ordinary `list`, indistinguishable from `if`/`let`/`when`. `SupportsCallGraph(ext)` reports this, so "does this language have a call graph" is answerable instead of folklore, and the docs table now has a Call graph column.

Elixir (#262) is the one gap I left open deliberately. It is homoiconic — `defmodule`, `def` and a real call are all the same `call` node — so wiring it would emit a reference to `"def"` per definition and attribute every call to a caller named `"def"`. It ships symbols-only with a test asserting zero call references, so the gap can't be closed by accident. Happy to do it properly in a follow-up if you want it.

## 4. Tests

`trace/extractor_refs_test.go` covers `ExtractReferences` for all 17 call-graph languages, asserting the `(callee, caller)` edge — the thing `trace callers`/`callees` actually consumes — plus the inverse for the symbols-only languages. Reverting Java, Rust and Python to the old node types fails it, so it is not vacuous.

## Smaller things

The elisp SHA discrepancy resolved in favour of the PR body: `elisp/parser.c` is byte-identical to `3a8f590`, and `29b4e49` only touched the Rust binding, so `elisp/README.md` was the wrong one. Corrected there.

Lint is clean on all three branches (`golangci-lint v2.13.2`, 0 issues), `go test -race ./...` passes.

On cost: I have no answer that makes 71 MB smaller, but the "compiled in but never reached" half of your objection is gone — every grammar is now enabled by default and 17 of 24 have a working call graph, with the other 7 documented as symbols-only rather than silently empty.

One caveat, same as on #264: I verified through unit tests that drive the real code paths, not by reproducing your end-to-end measurement with a live embedder. Worth a re-test on your side before you trust the numbers.

kryptt:
One addition to the above: pushing the fix surfaced a fifth problem, and it was the most serious of them.

**`.goreleaser.yml` still pinned `CGO_ENABLED=0`.** Your PR body note — "CGo is now in the default build path... Worth confirming the release matrix isn't surprised" — turned out to be the understatement. CI failed immediately:

```
imports github.com/yoanbernabeu/grepai/fsharp: build constraints exclude all Go files
```

The Build job cross-compiles five targets from one runner. With no target C toolchain, the Go tool doesn't error — it silently drops the CGO-only packages. Which means a tagged release from this branch would have shipped binaries with **no tree-sitter in them at all**, quietly, and the release job would have gone green.

Fixed:

- `ci.yml`: the Build job builds natively for linux/amd64 only. The Test job already builds and runs the full suite natively on ubuntu, macos and windows, so the cross-compile step was never adding coverage — only a toolchain requirement no runner meets. `CGO_ENABLED` is explicit now so a failure reads as a missing compiler rather than as excluded files.
- `.goreleaser.yml`: `CGO_ENABLED=1`, one build entry per target, each naming its own `CC`/`CXX`.
- `release.yml`: GoReleaser runs inside `ghcr.io/goreleaser/goreleaser-cross`, which bundles osxcross, aarch64 cross-gcc and mingw-w64.

I verified this one properly rather than assuming, since a broken release job is only discovered at tag time. `release --snapshot --clean` inside that image cross-compiles all five targets in 2m13s and produces the same five archives, checksums and homebrew formula as before, with unchanged names. `file` confirms Mach-O x86_64/arm64, ELF x86-64/aarch64, PE32+ x86-64, and the linux/amd64 binary runs and reports the new trace flags.

**One deliberate loss, and it needs your call:** windows/arm64 is no longer produced — goreleaser-cross ships no arm64 mingw toolchain, so that target cannot be a CGO build. It was only ever produced because CGO was off, so it never contained tree-sitter either; but it was in your release assets and now won't be. If you'd rather keep a CGO-free windows/arm64 artifact alongside the rest, that's one extra build entry and I'll add it.

Running a real release is the one thing I can't do from here, so that's worth a tag on a throwaway repo before you cut the next version.

All three PRs are green: Build, Lint, Test on ubuntu/macos/windows, both codecov gates.

PR #261, opened 2026-05-29
feat(trace): tree-sitter symbol extraction for 16 new languages

kryptt:
**Depends on / stacks on top of #260** (which wires up `--mode` dispatch). This PR adds the actual language coverage.

## Problem

Even after the dispatch fix in #260, tree-sitter symbol extraction only covers the original 9 grammars that landed via #152 (F#) and earlier (Go, JS/JSX, TS/TSX, Python, PHP, C#). For everything else — Ruby, Rust, Java, Scala, C, C++, Bash, Lua, Kotlin, Swift, SQL, Protobuf, HCL, Elm, TOML, Emacs Lisp — symbol extraction falls back to regex. Regex is usable but loses precision (Go method receivers, Python class docstrings, Rust trait impls, …) and silently misses syntax it doesn't know about.

We've already paid the CGo + grammar-linkage cost via #260. The marginal cost of linking more grammars and writing per-language queries is small in code and reviewable as one slice of `LangSpec` entries.

### Single registry, one entry per language

`trace/languages.go` already holds the registry from #260. This PR turns it into the single source of truth for both grammar registration **and** symbol extraction:

```go
type LangSpec struct {
    Name        string
    Extensions  []string
    GetLanguage func() *sitter.Language
    Queries     []NamedQuery   // ← new
}

type NamedQuery struct {
    Kind  string  // free-form, propagates to Symbol.Kind
    Query string  // tree-sitter S-expression with a @name capture
}
```

Adding a language is:
1. Import the grammar package (smacker subpackage, or vendor at repo root).
2. Add one `LangSpec` to `treeSitterLanguages`.
3. Write `trace/queries_<lang>.go` with the S-expression queries.
4. Drop a fixture in `trace/testdata/`.

No edits to `extractor_ts.go`. No new method per language. The dispatch picks up the new entry automatically.

### New query-based extraction path

`TreeSitterExtractor.extractViaQueries` runs each `NamedQuery` against the parsed tree using `sitter.NewQuery` + `NewQueryCursor`. It applies `#eq?` / `#match?` predicates via `cursor.FilterPredicates`. Emits one `Symbol` per `@name` capture with the `Kind` taken verbatim from the corresponding `NamedQuery`.

The legacy nine languages keep their hand-walked `extractGoSymbol` / `extractPythonSymbol` / … — `Queries` is empty on their `LangSpec` and the extractor falls through to the existing walk. **Zero behavior change for the original languages.**

### Vendored Emacs Lisp grammar

`smacker/go-tree-sitter` doesn't ship an elisp grammar, so we vendor [Wilfred/tree-sitter-elisp](https://github.com/Wilfred/tree-sitter-elisp) (MIT) at repo root in `elisp/`, mirroring the pattern set by `fsharp/`. Pinned at the last ABI-14-compatible upstream commit (`3a8f590`) since smacker's bindings top out at ABI 14. `elisp/README.md` documents the provenance + re-vendoring procedure.

Covered macros: `defun`, `defmacro`, `defvar`, `defconst`, `defcustom`, `defface`, `define-{minor,derived,generic,globalized-minor}-mode`, `cl-defun`, `cl-defmethod`, `cl-defgeneric`.

## Languages added (16)

| Tier | Languages | Extensions |
|---|---|---|
| **Top-8 priority** (rich queries: classes/traits/structs/enums/methods/...) | Ruby, Rust, Java, Scala | `.rb` `.rs` `.java` `.scala` `.sc` |
| **Medium** (functions + main types) | C, C++, Bash, Lua, Kotlin, Swift | `.c` `.h` `.cpp` `.cc` `.cxx` `.hpp` `.hh` `.hxx` `.sh` `.bash` `.zsh` `.lua` `.kt` `.kts` `.swift` |
| **Minimal** (top-level definitions) | SQL, Protobuf, HCL/Terraform, Elm, TOML | `.sql` `.proto` `.hcl` `.tf` `.elm` `.toml` |
| **Vendored** (top-8 priority) | Emacs Lisp | `.el` |

## Test plan

New `TestExtractor_LanguageFixtures` is a table-driven test that exercises every language end-to-end:

\`\`\`go
{name: "ruby", fixturePath: "ruby.rb", want: []string{
    "class:Hello", "class:Standalone", "method:say",
    "module:Greeter", "singleton_method:banner",
}},
\`\`\`

Each fixture has ~5 known symbols asserted set-wise — extras allowed because grammars sometimes surface more than the bare minimum. **16 sub-tests, all passing.**

The PR-1 `compound_test.go` cases that used `.lua` as the canonical regex-only extension are updated to `.zig` (still regex-only after this PR — no smacker Zig grammar).

\`go test ./... -count=1\` is clean. \`go vet ./...\` is clean.

## Numbers

- Binary size: 48 MB (post-PR-#260) → ~70 MB. Each grammar is roughly 1–3 MB compiled.
- Lines: +7,700 / -57 across 43 files. ~6,500 lines of that is the generated \`elisp/parser.c\` from Wilfred's grammar.
- Hand-written code: ~600 lines (queries_*.go averaging 10–20 lines each + LangSpec entries + tests).

## What's intentionally not in this PR

- **defalias** (elisp) — its target is a `quote` node, not a bare symbol; needs query-time descent into quote nodes.
- **Elixir** `def` / `defp` / `defmodule` / `defmacro` — these parse as `(call ...)` whose discriminator is the leading identifier; would need a query predicate of the form \`(#eq? @kind "def")\`-style on a deeply-nested structure. Doable but noisy enough to defer.
- **YAML, CSS/SCSS, HTML, Markdown, Svelte** — selectors / headings / mapping keys are dubious as "symbols" in the navigation sense; the regex extractor doesn't cover them either, and the value is lower.
- **Erlang, R** — no smacker grammar for either at this version. Could vendor in a follow-up.
- **Field-level extraction for Java / Kotlin** — useful but two nodes deep; deferred to a follow-up to keep this PR focused on the main definition kinds.

## Compatibility

- Single-project mode (`grepai watch` without `--workspace`): behaviour for the original 9 languages unchanged. New languages get tree-sitter-quality symbols where regex was previously used.
- Workspace mode (`grepai watch --workspace …`): same — the existing dispatch in #260 picks up the new languages automatically via `HasTreeSitterGrammar`.
- `--mode fast` still forces regex for everything, including the new languages, as an escape hatch.


yoanbernabeu:
See the detailed review on #260, which applies to this PR as well (the stack is linear: #262 ⊃ #261 ⊃ #260).

Specific to this PR: the 16 new grammars are compiled into the binary, but `config.DefaultConfig().Trace.EnabledLanguages` was never extended, so **11 of the 16 are never extracted by `grepai watch`** (Ruby, Scala, Bash, Kotlin, Swift, SQL, Protobuf, HCL, Elm, TOML, Emacs Lisp). Reproduced with a `.rb` file: `trace callers format_message --json` returns no symbol.

And for the ones that *are* enabled, `walkNodeForReferences`/`findContainingFunction` don't know the new node kinds, so call graphs are broken (Java `method_invocation`) or degraded (Rust `function_item` → `<top-level>` placeholder). `extractor_lang_test.go` only tests `ExtractSymbols`, never `ExtractReferences`, which is why CI stays green.

Non-regression on the existing languages is fine — I verified Go/Python/JS give the same symbols as before, and the `LangSpec.Queries`-empty fallback keeps the historical path intact. Licensing on the vendored elisp code is correct (MIT, provenance documented); just reconcile the pinned commit between the PR body (`3a8f590`) and `elisp/README.md` (`29b4e49`).

Full list of what I'd need before merging is on #260.

PR #262, opened 2026-05-29
feat(trace): elixir + elisp defalias + java fields + kotlin properties

kryptt:
**Stacks on #261** (which added 16 new tree-sitter languages). This PR fills the four follow-up items the parent PR explicitly deferred — each is small and reviewable in isolation.

### Elixir (`.ex`, `.exs`)

Every Elixir definition is parsed by tree-sitter as a `(call ...)` whose leading identifier is `def` / `defp` / `defmacro` / `defmodule` / etc. Three queries plus `#match?` predicates discriminate the shapes:

\`\`\`text
defmodule Greeter do ...   → arguments > alias                  → Kind: module
def hello(name), do: ...   → arguments > call > identifier      → Kind: function
def hello, do: ...         → arguments > identifier             → Kind: function
\`\`\`

The module query also catches `defprotocol`, `defimpl`, `defstruct`, `defexception`, `defrecord`. The function query catches `def`, `defp`, `defmacro`, `defmacrop`.

### Emacs Lisp `defalias`

`(defalias 'old-name #'new-name)` parses with the alias target inside a `quote` node, not as a bare symbol. New query descends one level:

\`\`\`text
(list . (symbol) @kind . (quote (symbol) @name) (#eq? @kind "defalias"))
\`\`\`

Picks up both plain quote (`'`) and sharp quote (`#'`) variants.

### Java fields

`(field_declaration declarator: (variable_declarator name: (identifier) @name))`. One Symbol per declared name, since Java lets `int a, b, c;` declare three at once.

### Kotlin properties

`(property_declaration (variable_declaration (simple_identifier) @name))`. Covers `val`, `var`, and `const val` at both class- and top-level scope.

## Test plan

Fixtures + table assertions in `trace/testdata/sample.ex`, `sample.el`, `java.java`, `kotlin.kt`. The `want` lists are strict — extra symbols are allowed but the listed ones must appear, so the queries can't silently regress on a smacker grammar bump.

All 17 `TestExtractor_LanguageFixtures` sub-tests pass. Full `go test ./...` clean. `go vet` clean. Binary size unchanged from #261 (~70 MB) since no new grammars were linked — just new queries.

## Numbers

- 10 files changed: `+116 / -8`
- 4 new queries (Elixir module, Elixir function-with-parens, Elixir function-paren-less, elisp defalias, Java field, Kotlin property)
- 2 new fixtures (`sample.ex`, `sample.el` was already there but didn't have defalias asserted)


yoanbernabeu:
See the detailed review on #260, which applies to this PR as well (the stack is linear: #262 ⊃ #261 ⊃ #260, verified by commit SHA — merging this one merges all three).

The four additions here are each correct in isolation. I verified Java field extraction works: `trace callers MAX_RETRY --json` returns `"kind": "field"`.

But they don't close the structural gap. Elixir (`.ex`/`.exs`) and Kotlin (`.kt`) are still absent from `EnabledLanguages`, so they stay inert by default like the other long-tail languages. And `trace callers helperMethod` on Java — a language that *is* enabled by default and that this PR treats as top-priority — returns **zero callers** even though `run()` calls it in your own fixture, because Java's `method_invocation` node isn't in `walkNodeForReferences`.

That last point is the one I'd most like fixed: it shows the call-graph gap isn't limited to the exotic languages, it affects the headline ones too.

Full list of what I'd need before merging is on #260.

PR #264, opened 2026-05-29
fix(trace): invalidate symbol cache when the extractor version changes

kryptt:
## Problem

The symbol-extraction dedup in `runInitialScan` skips any file whose stored content hash matches the freshly-scanned hash:

\`\`\`go
// cli/watch.go (upstream main)
if existingHash, ok := symbolStore.GetFileContentHash(fileInfo.Path); ok && existingHash == fileInfo.Hash {
    continue
}
\`\`\`

That's right about *file content* but ignores a second axis: the extractor itself may have changed since the symbols were persisted. A release that:

- ships a new tree-sitter grammar (e.g. lands C++ symbol coverage where there was none),
- broadens the regex patterns in `patterns.go`,
- fixes a bug that was silently swallowing certain definitions,

will produce a different symbol set for the same file content. But the dedup never re-runs the extractor on files whose source didn't change across the upgrade. Result: the new extraction quality is completely invisible to existing users until they manually delete `symbols.gob` or touch every source file.

I noticed this debugging an emacs workspace where the elisp extractor was substantially upgraded — the binary was redeployed, the watcher restarted, and the symbol index stayed exactly as it was. Same shape across other languages: any change to `extractor_ts.go`'s `walkNodeForSymbols` switch or `patterns.go`'s regex tables benefits only files modified after the upgrade.

## Solution

Pair each persisted symbol entry with the extractor signature that produced it. The dedup then requires *both* the content hash and the extractor version to match before skipping.

### Interface change

\`\`\`go
type SymbolExtractor interface {
    // … existing methods …

    // Version returns an opaque token that changes whenever this
    // extractor's output for the same input would change.
    Version() string
}
\`\`\`

\`RegexExtractor\` returns \`\"regex-v1\"\`; \`TreeSitterExtractor\` returns \`\"treesitter-v1\"\`. The convention is documented inline: bump the constant manually when symbol output changes meaningfully. Future releases that add a language or fix an extractor bump to \`regex-v2\`, \`treesitter-v2-lua\`, etc.

### Storage change

\`GOBSymbolStore\` grows a parallel \`FileExtractorVersions map[string]string\` next to the existing \`FileContentHashes\`. A new \`SaveFileWithSignature\` writes both fingerprints atomically; \`GetFileExtractorVersion\` reads them back. The legacy \`SaveFileWithContentHash\` stays for backwards-compat — old call sites keep working but the dedup will treat their entries as \"version unknown\" and re-extract once.

\`gob\` decoding is backwards compatible: older symbols.gob files without the new field decode cleanly with \`FileExtractorVersions == nil\`. The dedup treats this as a cache miss → one-time re-extraction → cache updates to the current version. Subsequent scans skip cleanly.

### Dedup change

\`\`\`go
existingHash, hashOK := symbolStore.GetFileContentHash(fileInfo.Path)
existingVersion, versionOK := symbolStore.GetFileExtractorVersion(fileInfo.Path)
if hashOK && versionOK && existingHash == fileInfo.Hash && existingVersion == extractor.Version() {
    continue
}
\`\`\`

Same logic flows into the fsnotify single-file update path (\`handleFileEvent\`) so the cache stays consistent across both startup-scan and live-edit flows.

## Test plan

- Existing \`TestRunInitialScan_SkipsSymbolExtractionWhenContentHashMatches\` is updated to seed via \`SaveFileWithSignature\` (matching the new contract). The original \`SaveFileWithContentHash\` seed now models the legacy-cache case and would correctly trigger a re-extract.
- New \`TestRunInitialScan_ReExtractsWhenExtractorVersionDiffers\` seeds a stale \`\"regex-vOLD\"\` and asserts the sentinel symbol is invalidated + the source's real symbol is re-extracted.
- New \`TestRunInitialScan_ReExtractsWhenExtractorVersionIsMissing\` seeds via the legacy \`SaveFileWithContentHash\` (simulating an old gob file with no extractor-version metadata) and asserts the same invalidate-and-re-extract behaviour.
- Mock \`SymbolExtractor\` implementations in tests grow \`Version()\`.

\`go test ./...\` is clean across the entire module.

## Migration notes

- First scan after upgrade re-extracts every file that's still in the symbol store. For a workspace with, say, 100k files this is a one-time cost on the order of a few minutes (extraction is fast; no embedding involved). Subsequent scans dedup cleanly.
- Bump the version string in \`extractor.go\` / \`extractor_ts.go\` when shipping changes to symbol extraction that should invalidate cached output. The bump is the canonical \"this release wants symbol re-extraction\" signal.

### DIFF
--- a/cli/watch.go
+++ b/cli/watch.go
@@ -832,8 +832,14 @@ func runInitialScan(ctx context.Context, idx *indexer.Indexer, scanner *indexer.
 			continue
 		}
 
-		// Skip extraction when content hash matches what we already persisted.
-		if existingHash, ok := symbolStore.GetFileContentHash(fileInfo.Path); ok && existingHash == fileInfo.Hash {
+		// Skip extraction when BOTH the content hash AND the extractor
+		// version match what was persisted last time. Content alone isn't
+		// enough — a release that ships better extraction (new tree-sitter
+		// grammar, expanded regex patterns, bug-fixed query) needs to
+		// re-process unchanged files to surface the improved symbols.
+		existingHash, hashOK := symbolStore.GetFileContentHash(fileInfo.Path)
+		existingVersion, versionOK := symbolStore.GetFileExtractorVersion(fileInfo.Path)
+		if hashOK && versionOK && existingHash == fileInfo.Hash && existingVersion == extractor.Version() {
 			continue
 		}
 
@@ -842,7 +848,7 @@ func runInitialScan(ctx context.Context, idx *indexer.Indexer, scanner *indexer.
 			log.Printf("Warning: failed to extract symbols from %s: %v", fileInfo.Path, err)
 			continue
 		}
-		if err := symbolStore.SaveFileWithContentHash(ctx, fileInfo.Path, fileInfo.Hash, symbols, refs); err != nil {
+		if err := symbolStore.SaveFileWithSignature(ctx, fileInfo.Path, fileInfo.Hash, extractor.Version(), symbols, refs); err != nil {
 			log.Printf("Warning: failed to save symbols for %s: %v", fileInfo.Path, err)
 		}
 		symbolCount += len(symbols)
@@ -2130,7 +2136,7 @@ func handleFileEvent(ctx context.Context, idx *indexer.Indexer, scanner *indexer
 			symbols, refs, err := extractSymbolsWithFramework(ctx, extractor, fileInfo.Path, fileInfo.Content, processors...)
 			if err != nil {
 				log.Printf("Failed to extract symbols from %s: %v", event.Path, err)
-			} else if err := symbolStore.SaveFileWithContentHash(ctx, fileInfo.Path, fileInfo.Hash, symbols, refs); err != nil {
+			} else if err := symbolStore.SaveFileWithSignature(ctx, fileInfo.Path, fileInfo.Hash, extractor.Version(), symbols, refs); err != nil {
 				log.Printf("Failed to save symbols for %s: %v", event.Path, err)
 			} else {
 				log.Printf("Extracted %d symbols from %s", len(symbols), event.Path)
--- a/cli/watch.go
+++ b/cli/watch.go
@@ -815,11 +815,17 @@ func runInitialScan(ctx context.Context, idx *indexer.Indexer, scanner *indexer.
 			continue
 		}
 
-		// Skip files that are unchanged since the last index run and already tracked.
+		// Skip files that are unchanged since the last index run, already
+		// tracked, and extracted by the current extractor version. The
+		// mtime fast-path must also check the extractor signature —
+		// otherwise an upgrade that ships better extraction never
+		// re-processes files whose source didn't change.
 		if !lastIndexTime.IsZero() {
 			fileModTime := time.Unix(file.ModTime, 0)
 			if (fileModTime.Before(lastIndexTime) || fileModTime.Equal(lastIndexTime)) && symbolStore.IsFileIndexed(file.Path) {
-				continue
+				if v, ok := symbolStore.GetFileExtractorVersion(file.Path); ok && v == extractor.Version() {
+					continue
+				}
 			}
 		}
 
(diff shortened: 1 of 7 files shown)


yoanbernabeu:
Thanks for this PR — the problem it identifies is real and worth fixing: after an upgrade that ships better symbol extraction, unchanged files keep their stale symbols forever, and the improvement stays invisible.

However, I tested the branch end-to-end and **the fix is currently a no-op in exactly the scenario it targets**.

## What I measured

1. Indexed a small Go project with a `main` build → `Symbol index built: 8 symbols extracted`
2. Re-ran the watcher with a build of this branch, on the existing `symbols.gob` (i.e. the upgrade scenario, sources untouched)
3. Result: `Symbol index built: 0 symbols extracted`

Expected: 8 symbols re-extracted, since the persisted store predates `fileExtractorVersions` and `GetFileExtractorVersion` correctly returns `ok == false`.

## Why

There is an earlier skip in `runInitialScan` (`cli/watch.go`, around line 818) that runs *before* the new version check:

```go
// Skip files that are unchanged since the last index run and already tracked.
if !lastIndexTime.IsZero() {
    fileModTime := time.Unix(file.ModTime, 0)
    if (fileModTime.Before(lastIndexTime) || fileModTime.Equal(lastIndexTime)) && symbolStore.IsFileIndexed(file.Path) {
        continue   // <-- returns here
    }
}

fileInfo, err := scanner.ScanFile(file.Path)
...
existingVersion, versionOK := symbolStore.GetFileExtractorVersion(fileInfo.Path)  // never reached
if hashOK && versionOK && existingHash == fileInfo.Hash && existingVersion == extractor.Version() {
    continue
}
```

On an upgrade, source files are unchanged, so their `ModTime` is older than `lastIndexTime` and they are already tracked in the symbol store. Every file exits through that first `continue` and never reaches the extractor-version comparison.

The unit tests in this PR pass because they exercise the second block directly (seeding the store and calling with a zero/future `lastIndexTime`), so they never traverse the first filter.

## Suggested direction

Teach the first filter about the extractor version too, so the mtime fast-path only applies when the persisted signature still matches:

```go
if !lastIndexTime.IsZero() {
    fileModTime := time.Unix(file.ModTime, 0)
    if (fileModTime.Before(lastIndexTime) || fileModTime.Equal(lastIndexTime)) && symbolStore.IsFileIndexed(file.Path) {
        if v, ok := symbolStore.GetFileExtractorVersion(file.Path); ok && v == extractor.Version() {
            continue
        }
    }
}
```

A regression test that goes through `runInitialScan` with a realistic `lastIndexTime` (in the past, as a real upgrade would have) and a store seeded via the old `SaveFileWithContentHash` would have caught this.

Everything else in the PR looks good to me: `go build`, `go vet` and `go test -race ./...` all pass, the store changes are backward compatible (an old `symbols.gob` loads fine and `trace`/`search` keep working after the upgrade), and `SaveFileWithSignature` is a clean API.

Happy to take another look once the first filter accounts for the extractor version.

kryptt:
Thanks for testing this and your suggestions

Fixed in 0802959: the mtime fast-path now only skips when the persisted extractor signature still matches.

```go
if !lastIndexTime.IsZero() {
    fileModTime := time.Unix(file.ModTime, 0)
    if (fileModTime.Before(lastIndexTime) || fileModTime.Equal(lastIndexTime)) && symbolStore.IsFileIndexed(file.Path) {
        if v, ok := symbolStore.GetFileExtractorVersion(file.Path); ok && v == extractor.Version() {
            continue
        }
    }
}
```

Added `TestRunInitialScan_ReExtractsWhenMtimeFastPathHitsStaleVersion`, which goes through `runInitialScan` with a past `lastIndexTime` and a backdated source mtime, seeded via the legacy `SaveFileWithContentHash` — the shape of a real upgrade. It also asserts the current signature is persisted afterwards, so the second scan skips cleanly rather than re-extracting forever. The test fails against the previous commit.

`TestRunInitialScan_SkipsIndexedFileByLastIndexTime` now seeds via `SaveFileWithSignature`; it was encoding the old contract and broke.

PR #266, opened 2026-06-12
fix(embedder): prevent OpenAI 300k token limit errors in batch embedding

### LINKED ISSUE #214: [Bug]: OpenAI embedding batches exceed 300k token limit, no progress saved between retries
### Bug Description

Initial indexing fails on medium-to-large codebases when using the OpenAI provider. grepai sends embedding batches that exceed OpenAI's 300,000 token-per-request limit. There's no config option to control the embedding batch size, and none of the existing knobs (`parallelism`, `rpg_max_dirty_files_per_batch`, `chunking.size`) affect it.

There are also two secondary issues that compound the problem:

1. **Full rescan on every retry** — when the embed step fails, grepai re-scans the entire project from scratch instead of resuming from where it left off. This means it re-discovers all the same files, re-chunks them, and tries the same oversized batch again.

2. **Workspace projects share rate limits** — when running as a workspace with multiple projects, all projects compete for the same OpenAI TPM quota. One project's initial indexing can burn through the rate limit and cause the other project to fail with a 429 as well.

```
→ grepai watch
Error: initial indexing failed: failed to embed batches: OpenAI API error (status 400): Invalid 'input': maximum request size is 300000 tokens per request.
```

### Steps to Reproduce

1. `grepai init` with OpenAI provider (`text-embedding-3-small`)
2. Point it at a codebase with ~1,500-2,000 source files (mix of Go, TypeScript, TSX)
3. Run `grepai watch`
4. Initial scan completes (2,100 files scanned), embedding starts on ~10,000+ chunks
5. Fails with a 400 error on the batch size

I tried reducing `parallelism` to 1, `rpg_max_dirty_files_per_batch` down to 16, and `chunking.size` from 512 to 128 — none of it changed the behavior. Reducing chunk size actually made it worse since it produced more total chunks (52k vs ~10k).

### Expected Behavior

- grepai should split embedding API calls to stay under OpenAI's 300k token-per-request limit, regardless of how many chunks the codebase produces
- On retry after an embed failure, it should resume from where it got to rather than rescanning and re-chunking everything
- Workspace projects should coordinate their rate limit usage or at least back off gracefully on 429s without a full restart

### Actual Behavior

Each retry does a full rescan + re-chunk before attempting to embed again.

After exhausting all retries, `grepai watch` exits. Running it again fails instantly with the same 400 error — none of the progress from earlier attempts was persisted, so it's back to square one:

```
$ grepai watch
Error: initial indexing failed: failed to embed batches: OpenAI API error (status 400): Invalid 'input': maximum request size is 300000 tokens per request.
```

### grepai Version

grepai version 0.35.0

### Operating System

Linux

### Embedding Provider

OpenAI

### Relevant Logs

```shell
Error: initial indexing failed: failed to embed batches: OpenAI API error (status 400): Invalid 'input': maximum request size is 300000 tokens per request.
```

### Configuration

```yaml
version: 1
embedder:
    provider: openai
    model: text-embedding-3-small
    endpoint: https://api.openai.com/v1
    dimensions: 768
    parallelism: 4
store:
    backend: gob
chunking:
    size: 512
    overlap: 50
watch:
    debounce_ms: 500
    rpg_persist_interval_ms: 1000
    rpg_derived_debounce_ms: 300
    rpg_full_reconcile_interval_sec: 300
    rpg_max_dirty_files_per_batch: 128
search:
    boost:
        enabled: true
        penalties:
            - pattern: /tests/
              factor: 0.5
            - pattern: /test/
              factor: 0.5
            - pattern: __tests__
              factor: 0.5
            - pattern: _test.
              factor: 0.5
            - pattern: .test.
              factor: 0.5
            - pattern: .spec.
              factor: 0.5
            - pattern: test_
              factor: 0.5
            - pattern: /mocks/
              factor: 0.4
            - pattern: /mock/
              factor: 0.4
            - pattern: .mock.
              factor: 0.4
            - pattern: /fixtures/
              factor: 0.4
            - pattern: /testdata/
              factor: 0.4
            - pattern: /generated/
              factor: 0.4
            - pattern: .generated.
              factor: 0.4
            - pattern: .gen.
              factor: 0.4
            - pattern: .md
              factor: 0.6
            - pattern: /docs/
              factor: 0.6
        bonuses:
            - pattern: /src/
              factor: 1.1
            - pattern: /lib/
              factor: 1.1
            - pattern: /app/
              factor: 1.1
    hybrid:
        enabled: false
        k: 60
trace:
    mode: fast
    enabled_languages:
        - .go
        - .js
        - .ts
        - .jsx
        - .tsx
        - .py
        - .php
        - .lua
        - .c
        - .h
        - .cpp
        - .hpp
        - .cc
        - .cxx
        - .rs
        - .zig
        - .cs
        - .java
        - .fs
        - .fsx
        - .fsi
        - .pas
        - .dpr
    exclude_patterns:
        - '*_test.go'
        - '*.spec.ts'
        - '*.spec.js'
        - '*.test.ts'
        - '*.test.js'
        - __tests__/*
rpg:
    enabled: false
    feature_mode: local
    drift_threshold: 0.35
    max_traversal_depth: 3
    llm_provider: ollama
    llm_endpoint: http://localhost:11434/v1
    llm_timeout_ms: 8000
    feature_group_strategy: sample
update:
    check_on_startup: false
ignore:
    - .git
    - .grepai
    - node_modules
    - vendor
    - bin
    - dist
    - __pycache__
    - .venv
    - venv
    - .idea
    - .vscode
    - target
    - .zig-cache
    - zig-out
    - qdrant_storage
    - '*.pb.go'
    - '*.connect.go'
    - '*_templ.go'
    - '*.gen.ts'
    - web/src/lib/gen
    - orchestrator/frontend/src/lib/api/gen
    - .agents
    - .codex
```

### Checklist

tcc-andy-shaw:
## Summary

- Fixes batch token estimation that underestimated code/CSS/HTML token counts (~4 chars/token → ~3 chars/token), causing batches to exceed OpenAI's 300,000 token request limit
- Reduces `MaxBatchTokens` from 280k to 250k for additional safety margin
- Adds `BatchError` type that surfaces which files are in a failing batch, making it easy to identify problematic files when errors do occur

## Root Cause

The `EstimateTokens` function used `len(text) / 4` (~4 chars per token), which is reasonable for English prose but too optimistic for CSS/HTML/code with special characters and short identifiers (which tokenize at ~2.5-3 chars per token). This caused batches estimated at 280k tokens to actually contain 300k+ real tokens, hitting OpenAI's hard limit.

## Test plan

- [x] All existing embedder and indexer tests pass
- [x] Tested on a large Angular project with heavy CSS/font-awesome content that previously failed
- [ ] Verify no regressions on projects that previously indexed successfully (batches will be slightly smaller, requiring more API calls)

Closes #214

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/embedder/batch.go
+++ b/embedder/batch.go
@@ -5,16 +5,15 @@ package embedder
 const MaxBatchSize = 2000
 
 // MaxBatchTokens is the maximum total tokens per OpenAI embedding API batch.
-// OpenAI has a 300,000 token limit. We use 280,000 for safety margin.
-const MaxBatchTokens = 280000
+// OpenAI has a 300,000 token limit. We use 250,000 for safety margin since
+// token estimation is approximate.
+const MaxBatchTokens = 250000
 
 // EstimateTokens estimates the token count for a text string.
-// Uses a conservative estimate of ~4 characters per token for English text.
-// This is intentionally conservative to avoid hitting API limits.
+// Code/CSS/HTML tokenizes at ~2.5-3 chars per token due to special characters
+// and short identifiers. We use 3 chars/token to avoid exceeding API limits.
 func EstimateTokens(text string) int {
-	// Rough estimate: 1 token ≈ 4 characters for English text
-	// Use 3.5 to be more conservative
-	return (len(text) + 3) / 4 // Round up
+	return (len(text) + 2) / 3
 }
 
 // BatchEntry represents a single chunk with metadata for tracking its source.
--- a/embedder/batch_test.go
+++ b/embedder/batch_test.go
@@ -392,10 +392,10 @@ func TestEstimateTokens(t *testing.T) {
 		{"a", 1},
 		{"ab", 1},
 		{"abc", 1},
-		{"abcd", 1},
+		{"abcd", 2},                     // 4 chars -> 2 tokens (ceil(4/3))
 		{"abcde", 2},                    // 5 chars -> 2 tokens
-		{"hello world", 3},              // 11 chars -> 3 tokens
-		{string(make([]byte, 100)), 25}, // 100 chars -> 25 tokens
+		{"hello world", 4},              // 11 chars -> 4 tokens (ceil(11/3))
+		{string(make([]byte, 100)), 34}, // 100 chars -> 34 tokens (ceil(100/3))
 	}
 
 	for _, tt := range tests {
@@ -408,8 +408,8 @@ func TestEstimateTokens(t *testing.T) {
 
 func TestFormBatches_TokenLimit(t *testing.T) {
 	// Create chunks that are large enough to trigger token limit
-	// Each chunk will be ~10000 chars -> ~2500 tokens
-	// With MaxBatchTokens = 280000, we can fit ~112 such chunks per batch
+	// Each chunk will be ~10000 chars -> ~3334 tokens (10000/3)
+	// With MaxBatchTokens = 250000, we can fit ~75 such chunks per batch
 	largeChunk := string(make([]byte, 10000))
 
 	// Create 200 large chunks - should be split into multiple batches by token limit
--- a/embedder/errors.go
+++ b/embedder/errors.go
@@ -33,6 +33,29 @@ func NewContextLengthError(chunkIndex, estimatedTokens, maxTokens int, message s
 	}
 }
 
+// BatchError wraps an embedding error with the file indices that were in the failing batch.
+type BatchError struct {
+	FileIndices []int
+	Err         error
+}
+
+func (e *BatchError) Error() string {
+	return e.Err.Error()
+}
+
+func (e *BatchError) Unwrap() error {
+	return e.Err
+}
+
+// AsBatchError extracts a BatchError from an error chain.
+func AsBatchError(err error) *BatchError {
+	var batchErr *BatchError
+	if errors.As(err, &batchErr) {
+		return batchErr
+	}
+	return nil
+}
+
 // IsContextLengthError checks if an error is or wraps a ContextLengthError.
 func IsContextLengthError(err error) bool {
 	var ctxErr *ContextLengthError
--- a/embedder/openai.go
+++ b/embedder/openai.go
@@ -8,6 +8,7 @@ import (
 	"io"
 	"net/http"
 	"os"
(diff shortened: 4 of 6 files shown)


yoanbernabeu:
Thanks for this — the analysis is right, and the `BatchError` type in particular is a genuinely nice idea.

I went with #226 for the fix itself. The two PRs attack the same symptom from different ends: yours makes the estimate more conservative (≈3 chars/token instead of 4, `MaxBatchTokens` 280k → 250k), #226 catches the API's own 400 "maximum request size" and splits the batch recursively until it succeeds.

The reason I took the reactive one is that better estimation lowers the probability without closing the hole. `len(text)/3` is a good heuristic, but content that tokenises below 3 chars/token — dense minified code, CJK, emoji-heavy text — still overshoots, and when it does the user gets the same hard failure as before, just more rarely. Splitting on the actual error is bounded by what the API says rather than by how well we guessed. I verified the split works (a batch of 8 walks 8 → 4 → 2 until every sub-batch succeeds, embeddings reassembled in order), and that the 400 isn't marked retryable, so there's no infinite-retry risk.

There's also a cost consideration: lowering the ceiling to 250k and the ratio to 3 shrinks every batch for every project, including the large majority that never hit the limit — more API round-trips and more latency across the board to avoid a rare failure. #226 leaves the normal path untouched and only pays extra calls in the case that actually needs it.

Two things from your PR I'd like to see land separately, if you're up for it:

- **`BatchError`** — surfacing which files are in the failing batch is a real diagnostic improvement, and it's orthogonal to how the limit is avoided. It would be just as useful on top of #226.
- **The token-estimation study** — your ~3 chars/token measurement on CSS/HTML/code is good data. Even with splitting in place, a less optimistic estimate means fewer splits and fewer wasted round-trips. It's a good optimisation on its own merits; it just shouldn't be the only line of defence.

Sorry to close rather than merge both — #226 and this one overlap enough in `embedder/openai.go` and `embedder/batch.go` that taking both as-is would have meant a messy manual resolution. A follow-up PR with just `BatchError` (and optionally the estimation change) would go in cleanly and I'd be glad to review it.

Issue #178, opened 2026-03-07
GOB index corruption on unclean shutdown — crash loop with 'unexpected EOF'


coolmanns:
## Description

When the system shuts down uncleanly (reboot, OOM kill, signal) while `grepai watch` is running, `index.gob` gets truncated. On restart, grepai fails with:

```
[ERROR] failed to load index: failed to decode index: unexpected EOF
```

This triggers an infinite crash loop since the corrupted file persists and grepai has no recovery path.

## Root Cause

`index.gob` appears to be written directly (non-atomically). If the process is interrupted mid-write, the file is left in a partial/truncated state (e.g., 384 bytes instead of the expected ~14-200MB).

## Impact

In my case, a server reboot on March 4 corrupted the index. grepai crash-looped for **3 days** (15,000+ restart attempts) before I noticed and manually deleted the index file.

## Suggested Fix

Use atomic writes for `index.gob`:
1. Write to a temporary file (`index.gob.tmp`)
2. `fsync` the temp file
3. Rename `index.gob.tmp` → `index.gob` (atomic on POSIX)

Alternatively, detect the corruption on startup and automatically rebuild instead of exiting with an error.

## Workaround

I wrote a systemd `ExecStartPre` script that checks `index.gob` size (<1KB = corrupted) and deletes it before grepai starts:

```bash
#!/bin/bash
INDEX="/path/to/.grepai/index.gob"
MIN_SIZE=1024
if [ -f "$INDEX" ]; then
    size=$(stat -c%s "$INDEX" 2>/dev/null || echo 0)
    if [ "$size" -lt "$MIN_SIZE" ]; then
        echo "index.gob is ${size} bytes — likely corrupted. Removing."
        rm -f "$INDEX" "$INDEX.lock"
    fi
fi
```

## Environment

- grepai v0.9.x (latest as of March 2026)
- Backend: GOB
- OS: Ubuntu Linux 6.17
- Service manager: systemd (user unit)

PR #271, opened 2026-07-02
feat(agents): grepai-doctor skill — self-service install/heal/benchmark for AI agents

rusel95:
## What

An **agent skill** for grepai: `.claude/skills/grepai-doctor/` (next to the repo's existing `grepai` and `release` skills) — a `SKILL.md` (the instructions AI coding agents auto-discover) plus one idempotent `doctor.sh` (~100 lines, bash, no daemons/launchd/root). Together they give any agent — Claude Code, Codex, Windsurf, anything that reads the emerging `.claude/skills/` layout — auto-loaded by Claude Code when working in this repo, and (discoverable by `npx skills add yoanbernabeu/grepai --skill grepai-doctor` once merged, verified against the skills.sh CLI discovery logic) — a single self-service entry point for the whole grepai lifecycle:

| mode | does |
|---|---|
| `install` | grepai (Homebrew tap or `install.sh`) + Ollama + `nomic-embed-text` |
| `init` | end-to-end setup of a fresh repo (runs `install` first, then `grepai init --yes` + background watcher) |
| `ensure` (default) | if the watcher is down: sweep truncated `index.gob` files (main repo **and** all linked worktrees), restart `grepai watch --background`; exits instantly when healthy — safe as a session-start hook |
| `force` | stop watcher, wipe indexes, full reindex — the escape hatch for a corrupted index |
| `bench` | measure `git grep` vs `grepai search` **on the host repo** (exact-symbol probe + natural-language queries, timings and hit counts), write `.grepai/bench.md`, append `grepai stats` |

## Why grepai should ship this

**1. It converts grepai's worst failure mode from "agent silently stops using grepai" into self-heal.** The README's core pitch is reducing AI-agent tokens — but when an agent hits `failed to decode index: unexpected EOF` (#178) or an empty index, it falls back to plain grep and *never comes back*. The user doesn't notice; grepai just stops earning its keep. A skill is exactly the mechanism agents already use to know what to do about a broken tool.

**2. It's the missing complement to `grepai agent-setup`.** `agent-setup` teaches agents *how to query*; this teaches them *how to install, repair, and justify* the tool. `bench` produces per-repo evidence (in our runs: natural-language queries — literal grep 0 hits vs grepai top-10 in ~0.1s; exact symbol `main` — grep 555 hits vs 10 focused chunks; `grepai stats` reporting ~96% token savings), which SKILL.md tells the agent to persist into its own memory — so the "use grepai here" decision survives across sessions instead of being re-litigated.

**3. Zero cost to the project.** Two static files; no Go code, no dependencies, no build/runtime impact, nothing to maintain in CI. If the convention doesn't interest you, closing this costs nothing — the skill also works fine distributed out-of-band. But shipping it in-repo means every clone gets it, and it stays in sync with grepai's actual CLI surface.

Field-tested on three repos (this one included: fresh `init` → 253 files indexed → bench → search verified). Cross-references issues #178/#211; degrades gracefully on versions with or without #269/#270.

## Notes

- The root README's Installation section gains an optional "AI agent skill" entry (`npx skills add yoanbernabeu/grepai --skill grepai-doctor`) next to the brew/install.sh methods.

- `shellcheck`-clean; the only mutations it performs live under `.grepai/` and the watcher lifecycle.
- Happy to move/rename the directory (e.g. `docs/agents/`, `contrib/`) or fold the how-to into the docs site instead if you prefer that shape.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


yoanbernabeu:
Thanks for putting this together — the underlying idea is genuinely good. A single self-service entry point for install/heal/benchmark is something this project could use, the `bench` mode (measuring grepai against plain grep instead of asserting it's better) is intellectually honest, and `doctor.sh` is careful in places: `set -u`, a `watcher_running` check before touching any index, BSD/GNU portability.

I'm closing it because of what the script is allowed to do without a human in the loop, not because of the idea.

`SKILL.md` instructs an agent to invoke this proactively whenever grepai looks broken. Combined with what the script does, that hands an autonomous agent the ability to:

- **install software** — `brew install yoanbernabeu/tap/grepai`, `brew install ollama`, or `curl -sSL .../install.sh | sh` (remote script piped to a shell), with no confirmation
- **start a persistent network daemon** — `nohup ollama serve >/dev/null 2>&1 &`, detached, outliving the session, outside `brew services` or launchd, listening on 11434 with no trace for the user to find later
- **download ~274 MB** (`ollama pull nomic-embed-text`) unprompted
- **delete indexes** — `force` removes `index.gob` with no size guard, in the main repo *and* every linked worktree

Any one of those is defensible when a human types it. Together, triggered implicitly by an agent's judgment that something is "broken", they're too much. The blast radius of a misfire is someone's machine, and `force` in particular is unrecoverable index loss behind a single keyword.

Two smaller things: the README entry points users at `npx skills add rusel95/grepai --skill grepai-doctor`, i.e. installing from an unreviewed fork; and `bench` issues real `grepai search` calls, which bills money if the configured provider is OpenAI/Requesty/OpenRouter, without saying so.

If you want to take another run at it, the shape I'd merge is: an explicit `--yes` (or an interactive prompt) gating `install` and `force`, `SKILL.md` telling agents to *report* what needs doing rather than to do it, no remote `curl | sh`, no detached daemon (defer to `brew services` or tell the user to start Ollama), and `bench` warning up front when the provider is a paid one. Read-only diagnostics — detect, explain, suggest — could stay fully automatic; that part is uncontroversial and useful on its own.

Sorry to close rather than leave it hanging, but I'd rather be explicit about the reason than let it sit. The `ensure` mode's truncated-index detection in particular is a good idea and overlaps with what #269 just landed on the read path — worth a look.

Issue #193, opened 2026-03-19
Feature request: manual workspace indexing + local-context incremental reindex for mcp-serve


gjed:
## Problem

When `grepai mcp-serve --workspace <name>` is configured as an MCP server in an AI coding agent (e.g., opencode, Claude Code, Cursor), a new `mcp-serve` process is spawned for **every session**. With multiple concurrent sessions and large workspaces, Ollama gets hammered with embedding requests and eventually becomes unresponsive.

**Setup:**
- 2 workspaces in `~/.grepai/workspace.yaml` (~280 projects total)
- Ollama in Docker (CPU-only, no GPU) serving `nomic-embed-text`
- Opening a new coding session spawns 2 `grepai mcp-serve` instances (one per workspace)
- With 4 concurrent sessions: 8 `mcp-serve` processes, all making embedding calls to Ollama
- Ollama pegged at ~400% CPU, eventually crashes

## Desired behavior

Two separate concerns that should be decoupled:

### 1. Workspace-level indexing: manual, explicit command

Full workspace indexing (all 280 projects) should be a deliberate action run once in a dedicated terminal, not triggered implicitly by opening sessions:

```bash
grepai watch --workspace gjed
```

`mcp-serve` should serve queries from the existing index without triggering any re-indexing of the full workspace.

### 2. Local context: incremental reindex of changed files only

When working inside a specific project that belongs to a workspace (e.g. opening opencode in `/home/marco/repos/gjed/grepai`), `mcp-serve` should only reindex the files that have actually changed in that **local context** — not the entire workspace.

This makes sessions cheap: instead of re-embedding hundreds of projects, only the handful of files modified since the last index run are processed.

## Expected workflow

```bash
# Once, in a dedicated terminal:
grepai watch --workspace gjed

# Each coding session starts fast:
# mcp-serve serves from existing index, only re-embeds locally changed files
```

## Environment

- grepai v0.35.0
- Ollama in Docker (CPU-only)
- ~280 projects across 2 workspaces

## Contributing

I'd like to contribute a fix for this. Before diving in, I'd appreciate some pointers from the maintainers on how this would integrate with the existing architecture — particularly around `mcp-serve` startup behavior and how local-context detection should interact with the workspace index.

PR #299, opened 2026-09-05
feat(cli): add one-time manual indexing

harumiWeb:
## Background

I use grepai for semantic search and code graph exploration across development projects.

In some projects, especially those with test suites or tooling that frequently create and delete temporary files, keeping `grepai watch` running continuously can interfere with those file operations. On Windows in particular, a watcher accessing a temporary file at the same time it is being removed or replaced can occasionally result in permission or file-locking errors.

For these projects, I would prefer not to keep the filesystem watcher running during normal development. Instead, I want to refresh the grepai index at explicit synchronization points, such as:

* before a Git commit
* after creating or checking out a worktree
* before running grepai-based analysis
* in CI environments

Until now, `grepai watch` has effectively been the entry point for both the initial indexing pass and continuous filesystem monitoring, so there was no straightforward way to perform only the initial indexing step and exit.

This PR adds a one-time `grepai index` command that reuses the existing indexing pipeline without starting the filesystem watcher. This makes it possible to keep the index up to date through Git hooks or other explicit workflows while avoiding a long-running watcher process.

The existing `grepai watch` workflow remains unchanged for users who prefer continuous indexing.

## Summary

- add `grepai index` to update the current project's vector, symbol, and optional RPG indexes once, without starting a watcher or daemon
- add `grepai index --workspace <name>` to index every project in a configured workspace, continue after per-project failures, and return an aggregate error
- share the initial indexing runtime with `grepai watch`, including incremental skip behavior, deleted-file cleanup, and RPG rebuilding
- keep persistence policy at the caller boundary: one-shot indexing returns persistence failures, while `watch` preserves its existing warning-and-continue behavior
- document manual indexing for pre-commit, linked-worktree, and CI workflows, including that pre-commit indexes the full working tree rather than only staged content

## Related Issue

Addresses the manual project/workspace indexing portion of #193. The `mcp-serve` local-context behavior requested there remains out of scope, so this PR does not close the issue.

## Tests

- `CC=gcc go test -race ./...`
- `CC=gcc go vet ./...`
- `go test ./cli`
- `go run ./cmd/grepai index --help`
- `go run ./cmd/gendocs/main.go`, followed by `npm run build` in `docs/`

`CC=gcc golangci-lint run ./cli/...` was also attempted. It remains blocked by pre-existing gofmt findings in untouched CLI files and the existing G118 finding in the watch supervisor.

A live-provider smoke test was not run because this checkout has no `.grepai/config.yaml`; the command lifecycle and no-watcher path are covered by the CLI tests.


yoanbernabeu:
Thanks for this, and sorry for the timing — #300, #303 and #306 landed on `main` today and this branch now conflicts, on `CHANGELOG.md` and `cli/watch.go`.

`cli/watch.go` is the real one: #306 rewrote a large part of the watcher startup path to fail closed when filesystem coverage is incomplete, which lands right where `newProjectIndexRuntime` factors the initialization out. The two refactors want the same lines, so the merge needs a judgement call rather than a mechanical resolution.

Could you rebase on current `main`? The extracted-runtime approach is a good idea and I would like to take it — I just want a clean base and green CI first.

One thing to know while rebasing: since #300, the GOB stores only persist when something actually changed, so a run that indexes nothing writes no `index.gob` at all. Worth checking that one-shot indexing still persists what you expect on a fresh project.

harumiWeb:
Thanks for the heads-up. Also, thank you for accepting this feature request. I’ve rebased the branch onto the current `main` and force-pushed the updated commits.

For the `cli/watch.go` conflict, I kept the startup fencing, readiness, abort, and fail-closed behavior introduced by #306, then reapplied the shared indexing runtime within that lifecycle.

I also aligned `grepai index` with the new project writer-lock behavior and added regression coverage for the GOB persistence changes from #300:

- a fresh project with indexable files creates and persists its vector and symbol indexes;
- a fresh empty project exits successfully without creating clean GOB files;
- an unchanged indexing run does not rewrite the existing GOB files.

`go test -race ./...`, `go vet ./...`, and the documentation build pass locally. The branch is now conflict-free and mergeable.

Thanks again for the guidance.

PR #300, opened 2026-09-05
fix(store): stop rewriting unchanged local indexes

### LINKED ISSUE #298: [Bug]: Idle watchers and read-only trace requests rewrite large index files
## The problem

grepai can repeatedly rewrite a large saved index even when the project is idle. Read-only trace requests from an AI assistant can cause the same writes, although they only ask about existing code.

In the reported setup, `.grepai/symbols.gob` was about **1.4 GB** and was rewritten roughly every 30 seconds. This produced large bursts of disk activity without source changes. On a shared machine, that unnecessary work competes with other applications for disk bandwidth.

## How to reproduce

1. Index a project using the default local-file storage and leave `grepai watch` running.
2. Stop editing source files and observe the saved index's modification time and disk activity across several 30-second intervals.
3. Separately, ask a read-only MCP trace question, such as which functions call a particular function, and observe the index file again.

In the affected build, these operations rewrite the index even though its data has not changed.

## Expected behavior

- An idle watcher leaves unchanged index files alone.
- A read-only trace request does not rewrite the index.
- A real source change is still saved successfully.
- A failed save can be retried without discarding the pending update or the last good saved index.

## Reported environment

grepai **0.35.0**, Linux, Ollama, local GOB storage, with RPG disabled. GOB is grepai's local binary index-file format.

The original measurements recorded five large write bursts over about two minutes, approximately 30 seconds apart.

PR #300 addresses unnecessary idle/read-only saves. The cost of full saves after genuine edits is a separate problem addressed by the optional PostgreSQL work in #301.

jeremyakers:
Fixes #298.

## The problem

grepai can repeatedly rewrite its saved index even when no source files have changed. Asking an AI assistant a read-only question such as “what calls this function?” can also trigger a save.

That is expensive on large projects. The report in #298 describes a `symbols.gob` file of about **1.4 GB** being rewritten roughly every 30 seconds. These unnecessary writes use disk bandwidth and can slow other applications using the same disk.

## What this PR changes

The local-file stores now remember whether their data has changed. If nothing changed, the regular save timer and read-only queries leave the index files alone. After a real change, grepai still saves the updated index. If that save fails, it keeps the update pending so it can try again.

This also protects saved data when multiple processes use the same project: only one watcher may write that project's index at a time, and a failed replacement must not destroy the last good file.

## What users get

- An idle watcher no longer repeatedly writes large, unchanged index files.
- Read-only trace queries no longer rewrite these index files.
- Failed saves and competing watchers are less likely to lose recently indexed data.

Existing local indexes and settings continue to work. The local format, called GOB, is unchanged, and no database is required. Storage for the separate repository-graph feature (RPG) is not changed by this PR.

**A real edit still requires a full local-file save.** The separate PostgreSQL option in #301 addresses that remaining cost for busy projects.

## How it was checked

- Left a real watcher idle, made a source edit, and verified that only the edit caused a save.
- Ran CLI and AI-assistant trace queries against a copied large index and verified that its contents and modification time stayed unchanged.
- Tested failed saves, competing writers, and safe file replacement; Linux, macOS, and Windows CI pass.


jeremyakers:
Codecov feedback reviewed on exact head `25e352b`. The configured patch/project checks pass. Additional regression coverage was added for dirty-state edge cases, missing-reader overwrite, atomic replacement failure, writer-lock lifecycle/contention, and worktree auto-init failure paths; vector/symbol store changed-line coverage is now ~96%. Remaining uncovered lines are primarily OS/filesystem failure branches. The report's “142 commits behind” warning refers to Codecov's stale baseline upload (`a322537`), not this PR: Git reports the PR 0 commits behind `origin/main` (`b355512`), 9 ahead, and GitHub reports it mergeable/CLEAN. Linux, macOS, Windows, lint, build, and both Codecov checks all pass.

PR #301, opened 2026-09-06
feat(trace): add PostgreSQL storage for code relationships

### LINKED ISSUE #309: [Feature]: Small code edits still rewrite the entire symbol index on large projects
## The problem

Small source edits can require large disk writes because grepai saves its code-relationship index as one local `symbols.gob` file. That index contains function/class names and the calls used by trace commands and AI-assistant code navigation.

On a large project, the file can be around **1.5 GB**. If developers or coding agents keep changing files, grepai repeatedly rewrites that entire file just to save the relationships for a few changed files.

## Why the idle-save fix is not enough

Issue #298 and PR #300 address saves that happen when nothing changed. This is different: an active project has real changes that need to be saved, but the amount written is much larger than the change itself.

Putting search data in PostgreSQL does not solve this, because the code-relationship index still uses the local file.

## Expected behavior

Offer a storage option that can save the relationships for a changed source file without rewriting every other file's relationships.

For users already running PostgreSQL, it should be possible to store both kinds of index there. Projects sharing a database must keep their results separate. Existing local data should be importable without losing the original on a failed import.

Simple local storage should remain the default; small projects should not be forced to run a database.

## Why it matters

This would reduce unnecessary whole-file writes during active development and make large, continuously updated projects more practical to run.

PR #301 provides the optional PostgreSQL implementation. This issue is related to #298, but does not replace the separate idle/read-only-save fix.

jeremyakers:
Fixes #309. Related to #298 and #300.

## The problem

grepai's code-relationship index records functions, classes, and calls between them. Trace commands and AI-assistant tools use it to answer questions such as “what calls this function?”

Today, that information is saved in one local `symbols.gob` file. On a large project it can grow to around **1.5 GB**. Avoiding unnecessary idle saves helps (#300), but a busy project still changes frequently: a small, legitimate edit can cause the entire file to be rewritten.

Using PostgreSQL for regular search does not solve this on its own, because the code-relationship index still uses the local file.

## What this PR changes

This adds an **optional PostgreSQL store for code relationships**. Instead of rewriting one large file, grepai replaces the database records belonging to the changed source file. Projects sharing a database keep their results separate.

Existing local data can be imported when the option is enabled. The original file is kept until the import succeeds, and an interrupted import can be retried. Storage failures stop initialization rather than making an incomplete index look ready.

The PR also removes unnecessary work from index-status requests: asking whether an index is ready no longer requires calculating unrelated storage-size and reference statistics.

## Why it matters

Large projects and continuously active coding agents can make small updates without repeatedly rewriting a huge relationship index. Users already running PostgreSQL for search can use it for trace data too.

## Compatibility

Local-file storage remains the default. PostgreSQL must be explicitly enabled with `trace.store_backend: postgres`; normal trace commands and AI-assistant tools keep the same interface. This does not change how search data is stored or add MySQL support.

## How it was checked

- Imported a copied large index and tested interrupted imports, retries, and concurrent updates with real PostgreSQL.
- Checked CLI and AI-assistant queries, including isolation between projects.
- Local tests and Linux, macOS, Windows, and PostgreSQL CI pass.


yoanbernabeu:
Heads-up: #300 and #306 are now on `main`, and this branch conflicts with them — `CHANGELOG.md` and `cli/watch.go` (5 blocks).

The substantive one is `runWatchLoop`: this branch widens `symbolStore` to the `trace.SymbolStore` interface so Postgres can be swapped in, while #306 changed the same signature to take `w watchSource`. Both changes are wanted, they just need to be combined by someone who knows what each was for. Could you rebase on current `main`?

What landed:
- #300 — GOB stores persist only when dirty, plus a per-project writer lock; `watchProjectWithEventObserver` acquires it and passes the canonical root down.
- #306 — fail-closed watcher coverage; `onFatal` added to `watchProjectWithEventObserver`, body moved to `watchProjectWithEventObserverAndFence`.

No concerns with the approach here — opt-in `trace.store_backend: postgres` with GOB staying the default is exactly right. I just want a clean base and green CI before merging.

jeremyakers:
Updated this branch with current main while preserving the existing history. The merge conflicts are resolved, and the reviewed PostgreSQL fixes now cover safe schema initialization/legacy adoption, bounded migration waits and buffers, coherent trace/reference reads, and mutation freshness.

The refreshed head passed fork Code Review, the full local PostgreSQL race/shuffle suite, actual CLI/MCP checks, and upstream CI on Linux, macOS, Windows and PostgreSQL. The combined tree with #307/#308 also passed local tests and real usage checks.

Correction to my earlier merge-state note: there is no repository-side merge blocker. #301 modifies `.github/workflows/ci.yml`, while my GitHub CLI OAuth token lacks the separate `workflow` scope; GitHub reports `BLOCKED` only through that token. The public API reports `mergeable_state: clean`, matching the website. This is the documented GitHub CLI behavior in cli/cli#9145, not an unmet review/check/ruleset gate. The workflow change is intentional: it supplies PostgreSQL to integration tests and runs the trace, MCP, and CLI database-backed suites.

Older watcher behavior gaps remain separately tracked in #311, #312 and #313; they are not claimed as fixed here.

PR #306, opened 2026-09-07
fix(watch): stop with a clear error when file watching fails

### LINKED ISSUE #304: [Bug]: Watcher reports ready after it stops receiving changes from some folders
## The problem

grepai can report that it is ready even after it stops receiving changes from some project folders. The process stays alive, so users may continue trusting search and trace results that are no longer being updated.

The observed Linux warning was:

```text
Failed to watch .../src/log: no space left on device
```

**The disk was not full.** Linux limits how many file watches a user can hold across all applications. The system had reached 65,536 of 65,536 watches. grepai was using 4,919; another process had consumed the remaining allowance.

## What happened

New directories could not be watched. grepai logged the errors, continued reporting ready, and did not receive later changes from those directories. Freeing the watch allowance and restarting grepai restored coverage.

This can happen during startup or after a watcher has been running successfully for some time.

## Expected behavior

- Do not report ready when required folders could not be watched.
- If watching fails later, report a clear failure instead of continuing silently with stale results.
- Identify the affected path and explain that this particular error is a file-watch limit, not a disk-space problem.

## Reproduction and environment

Observed on Linux with a main-based grepai 0.36.x development build in a three-project workspace. The failure is not specific to its storage backend.

In an **isolated test environment**, reach the file-watch limit, start a watcher or create a new project folder, and check whether grepai wrongly remains ready after reporting the error. Do not exhaust a normal workstation's shared watch allowance just to reproduce this.

Nested projects were checked and were not registering duplicate directory watches in this incident. PR #306 addresses the misleading ready state; changing limits or reducing watch usage is separate work.

jeremyakers:
Fixes #304.

## The problem

A running grepai process is not necessarily receiving every file change. If the operating system refuses a directory watch, grepai previously logged a warning but continued to report that it was ready. Changes in the affected folder could then be missed indefinitely.

The reported Linux incident reached the user's shared file-watch limit. The error said **“no space left on device,” but the disk was not full**. Another application had used the remaining watch allowance.

## What this PR changes

- During startup, grepai reports ready only after it can watch all required projects and directories. Otherwise it stops with a clear error.
- If watching fails later, or the watched project root is removed or renamed, grepai stops rather than silently continuing with incomplete results.
- The error identifies the affected path. Linux watch-limit errors explain what limit was reached and what the user can do about it.
- The same rules apply across a workspace and additional Git working copies, including copies discovered while grepai is running.

On failure, grepai prevents new indexing work and lets already-started work stop before removing its ready status. It avoids an extra shutdown save that could make an incomplete update look final. **Normal Ctrl+C or `watch --stop` still saves normally.**

## Why it matters

Users and AI assistants should not trust stale results simply because the process is still alive. A visible failure makes the problem discoverable and gives the user a clear reason to restart after fixing it.

This does not raise operating-system limits or add duplicate-watch detection. It also cannot force an unresponsive storage operation to finish instantly. Deleting or moving a subfolder while the rest of the project remains available is handled separately by #307.

## How it was checked

Tested startup and runtime watch-limit failures in an isolated account, real project-root moves, normal shutdown, and multi-project operation. Automated tests cover failures during startup and saving; Linux, macOS, and Windows CI pass.


jeremyakers:
Advanced to 0c6a1747d474d1f28eb8fd3938e4272245ad91ba after fork-first review https://github.com/jeremyakers/grepai/pull/7 (current-head Codex code/security completed without findings), full race/shuffle/static validation, exact-commit Oracle approval, and real CLI/fsnotify SIT. The follow-up fences dynamic linked startup through registration and persistence-bearing teardown, including graceful peer persistence, before fatal readiness withdrawal. Real root-loss canceled a linked startup blocked in a loopback embedding request; the same PID exited nonzero with markers removed and no post-withdraw index changes. A graceful RPG-enabled control persisted valid vector/symbol/RPG indexes and exited zero. No installed soak binary has changed yet.

jeremyakers:
Advanced to cb15d57a201e022ffbc0a89b4bbd48ce99522336 after fork PR #8 current-head Codex code/security completed without findings: https://github.com/jeremyakers/grepai/pull/8. Workspace graceful persistence and all persistence-bearing deferred store closes now use the shared fence, and early store cancellation preserves the original peer fatal cause. Full race/shuffle/static checks, exact Oracle review, and actual two-project CLI workspace PGvector/GOBsymbol/RPG graceful plus real root-loss SIT passed. Root-loss process exited 1 with the native root-removed/renamed error and withdrawn markers; state digests stayed unchanged afterward. In-flight ordering is covered by deterministic race/barrier tests. The soak installation is still unchanged pending aggregate validation.

Issue #195, opened 2026-03-21
Qdrant stats/file listing truncate after one scroll page in large collections


garitar:
## Summary

`GetStats()` / file-list-style status paths against the Qdrant backend only account for the first scroll page, so large collections can report incomplete file counts and incomplete file listings.

## Problem

The current logic uses a single scroll page when deriving file-level stats from Qdrant payloads. On collections larger than that page size, the operator-facing counts/listings are truncated even though the underlying collection contains more points.

## Expected behavior

Stats and file listings should paginate through the full result set before computing distinct file counts or returning file-level entries.

## Why this matters

This creates misleading status output on larger indexes and makes Qdrant-backed operational introspection less trustworthy than the actual stored index state.

## Reproduction shape

Any Qdrant-backed project with more points than one scroll page can hit this. The exact threshold depends on the configured page size, but the problem is structural rather than dataset-specific.

## Related context

This is separate from:
- nested linked-worktree initialization behavior in [Issue #189](https://github.com/yoanbernabeu/grepai/issues/189)
- nested linked-worktree status/accounting behavior in [Issue #192](https://github.com/yoanbernabeu/grepai/issues/192)
- MCP startup/indexing behavior in [Issue #193](https://github.com/yoanbernabeu/grepai/issues/193)

PR #307, opened 2026-09-09
fix(watch): keep results current when folders change

### LINKED ISSUE #310: [Bug]: Folder moves, deletions, and ignore changes leave live search results out of date
## The problem

Folder-level changes are not always reflected completely while `grepai watch` remains running:

- A folder moved out of or deleted from the project can leave old code in search and trace results.
- A populated folder moved into the project may have files missing from the index until they are edited or grepai restarts.
- Changes to ignore rules can leave newly ignored files in results, or fail to discover files that have become allowed again.

## How to reproduce

Using a configured project with an initial index:

1. Start `grepai watch` and confirm that a function inside a subfolder can be found with search or trace.
2. Move that subfolder outside the project while leaving the watcher running. Repeat the query and check for stale results.
3. Move a different folder containing supported source files into the project. Check whether its existing files become searchable without editing each file or restarting grepai.
4. Change an ignore rule and check that both newly ignored and newly allowed files are handled.

## Expected behavior

The running watcher should update all affected index entries, not just the folder name or a subset of its files. It should respect configured file extensions and ignore rules, and keep up with large imports without silently dropping changes.

**Only grepai's saved index should be changed. Cleanup must never delete the user's source files.**

## Why it matters

Project reorganizations and generated folders are common in development. Search and trace tools should describe the current code, not a mixture of removed and undiscovered files.

PR #307 addresses this live-update problem. Losing the entire watched project is covered by #304/#306; changes made while grepai is stopped are handled by the startup work in #308. Issue #195 is related to complete Qdrant file listing but also requests separate statistics work.

jeremyakers:
Fixes #310.

## The problem

Folder changes can leave grepai's results out of date while the watcher is still running. Moving a folder out of a project can leave its old code in search and trace results. Moving in a folder that already contains files may leave some of that code missing until it is edited or grepai restarts.

Changing ignore rules can cause a similar mismatch: files that should disappear remain indexed, or newly allowed files are not discovered.

## What this PR changes

- Removes old index entries for files in a deleted or moved-out folder.
- Finds and indexes supported, non-ignored files already inside a folder moved into the project.
- Applies ignore-rule changes while running, including discovering and watching files that become allowed again.
- Honors configured file extensions and handles large imports without silently dropping changes when indexing falls behind.
- Limits cleanup to the affected folder: changing `src` does not remove results from `src-old`.

**Cleanup removes grepai's saved search/trace entries, not the user's files.** Ordinary editor saves that replace a file continue to work.

## Why it matters

Developers and coding agents regularly reorganize projects without restarting every tool. Search should show the code that is actually there, and trace results should not point to files that have left the project.

## Scope

This handles changes while the watcher is running. Restart-time cleanup is covered by #308; losing the entire watched project is covered by #306. With local-file storage, other processes see index updates after the normal scheduled save, not necessarily instantly.

Related to #195: cleanup now reads the complete Qdrant file list. The separate statistics problem in that issue is not claimed as fixed.

## How it was checked

Kept the same watcher running while testing file/folder removal, updated trace results, an import of more than 300 files, custom extensions, ignore changes, and folder replacements. Linux, macOS, and Windows CI pass.


yoanbernabeu:
Heads-up: #300 and #306 are now on `main`, and this branch conflicts with them.

I tried the merge locally and stopped: `watcher/watcher.go` alone has 6 conflicting blocks, and they are not mechanical. #306 restructured `Close()` into `Abort()`/`stop()`/`stopped()` with `RegistrationError` for fail-closed coverage, while this branch restructures `Close()` and `processEvents` for directory reconciliation. `cli/watch.go` and `trace/store.go` conflict too.

Since you wrote both, you are far better placed than me to arbitrate how the two shapes combine — I would rather not guess at the intent in the watcher core. Could you rebase this branch on current `main`?

For reference, what landed:
- #300 — GOB stores persist only when dirty, plus a per-project writer lock (`fileutil.AcquireProjectWriterLock`) that `watchProjectWithEventObserver` now takes before loading stores, handing the canonical root down.
- #306 — fail-closed watcher coverage, with `onFatal` added to `watchProjectWithEventObserver` and the mutation fence in `watchProjectWithEventObserverAndFence`.

One thing worth knowing while you rebase, since it bit me on #306: with #300 in place, startup persistence writes nothing when the initial scan produced no content, so a test that assumes `index.gob` exists after scanning an empty project will now fail. Giving the fixture project a real source file fixes it.

The review itself looks good from here — I just want a clean base before merging.

jeremyakers:
Updated this branch with current main using a normal merge, so the existing history is preserved. The integration keeps the fail-closed watcher lifecycle and writer lock, and keeps ordinary cancellation from closing the error channel before the CLI finishes saving.

The full race/shuffle suite, vet/lint, and live directory/file-deletion CLI tests passed. CI on the refreshed head is green on Linux, macOS, and Windows, and GitHub now reports the PR as mergeable.

Issue #218, opened 2026-04-07
[Bug]: Background watcher fails on large projects: 30s health check timeout too short


vbourdeix:
## Description

  When running `grepai watch --background` on a large project (~1717 files, Angular/TypeScript), the background watcher always fails with:

  Error: timeout waiting for process to become ready after 30s

  The initial scan + embedding phase takes significantly longer than 30s, so the health check kills the process before indexing completes.

  ## Environment

  - grepai v0.35.0
  - macOS (Apple Silicon)
  - Ollama with nomic-embed-text (responding in ~300ms per embedding)
  - Project: ~1717 source files (TypeScript/Angular)

  When running in foreground, the indexing does progress (files get scanned and embedded) but never completes before the background health check kills it.

### Steps to Reproduce

1. `grepai init` in a project with ~1500+ files
2. `grepai watch --background`
3. Process starts scanning/embedding but gets killed after 30s health check timeout

 ## Workaround attempts

  - Running in foreground (`grepai watch --no-ui`) works but the process gets killed when the terminal closes
  - Reduced `parallelism` to 1 in config — doesn't help because the 30s timeout is on the health check, not on Ollama
  - With default parallelism (0/unlimited), Ollama gets saturated and most embeddings fail with `context canceled`

### Expected Behavior

The `--background` mode should either:
  - Have a configurable health check timeout (e.g. `--timeout 300`)
  - Or consider the process "ready" as soon as it starts watching, not after the initial scan completes
  - Or perform the initial scan asynchronously after reporting readiness

### Actual Behavior

The --background mode health check kills the watcher process after 30s if the initial scan hasn't completed. On a ~1717 file project with Ollama (nomic-embed-text, ~300ms per embedding), the scan needs  
  several minutes. The process is terminated before any file gets persisted to the index, resulting in Files indexed: 0 every time. Running in foreground shows the indexing does progress normally, confirming
   the issue is specifically with the background health check timeout.    

### grepai Version

v0.35.0

### Operating System

macOS

### Embedding Provider

Ollama

### Relevant Logs

```shell
[grepai-watch] Starting grepai watch in /path/to/project
  [grepai-watch] Provider: ollama (nomic-embed-text)
  [grepai-watch] Backend: gob
  [grepai-watch] Performing initial scan...
  [grepai-watch] [STOPPED] /path/to/project - shutdown timeout
```

### Configuration

```yaml

```

### Checklist

- [x] I have searched existing issues to avoid duplicates
[...]

Issue #11, opened 2026-01-09
[Feature]: Add pre-commit Makefile target for local checks


yoanbernabeu:
## Problem Statement

Contributors may push code that fails CI checks (lint, formatting, tests) because they don't run all verifications locally before committing. This leads to:
- Wasted CI minutes
- Back-and-forth commits to fix formatting issues
- Delayed PR reviews (as seen in PR #10 with a simple `gofmt` issue)

Currently, contributors need to manually run multiple commands (`make lint`, `make test`, `gofmt -w .`, `go vet ./...`) which is easy to forget.

## Proposed Solution

Add a single `make pre-commit` target that runs all checks in sequence:

```makefile
pre-commit:
	gofmt -w .
	go vet ./...
	golangci-lint run ./...
	go test -race ./...
	@echo "✓ Ready to commit!"
```

This would:
1. **Format** all Go files automatically
2. **Vet** - detect common errors
3. **Lint** - run golangci-lint (comprehensive checks)
4. **Test** - run tests with race detection

Usage:
```bash
make pre-commit
# If all passes, safe to commit/push
```

## Alternatives Considered

- **Pre-commit hooks (Husky/Lefthook)**: More intrusive, requires setup from each contributor
- **Only rely on CI**: Current approach, but wastes time waiting for CI feedback

## Category

CLI / Commands

## Additional Context

This came up during review of PR #10 where a simple `gofmt` formatting issue was only caught after pushing to CI. A local `make pre-commit` would have caught this instantly.

PR #308, opened 2026-09-09
perf(indexer): reuse unchanged files when restarting the watcher

### LINKED ISSUE #216: feat(indexer): use per-file ModTime to skip unchanged files on watch restart

jeremyakers:
Fixes #216.

## The problem

Restarting `grepai watch` can repeat expensive work for files that have not changed. The old shortcut relies on one “last indexed” time for the whole project, rather than remembering what was checked for each file. A file edited during a long scan can also fall on the wrong side of that cutoff and be treated as already covered on the next restart.

This is especially costly for large projects and frequent branch switches: users wait for startup and may repeat requests to the embedding model—the model that prepares code for semantic search.

## What this PR changes

grepai now checks each file against the modification time saved with its own completed index entry.

- If that time still matches, it can reuse the saved result without rereading and re-embedding the file.
- If the time changed but the contents did not, it updates the saved time without asking the embedding model to do the same work again.
- Old index entries without reliable per-file times are checked before the shortcut is used.
- Previously indexed files reached through linked directories also reuse their unchanged saved records. Recovery no longer re-embeds them just because the normal folder walk did not visit them.
- Before reusing a recovered record, grepai checks that its referenced search chunks are actually present and belong to the right file. Older workspace references and Windows path forms remain readable; identifiers for chunks split by the model are preserved in full.
- Startup also checks saved entries against the current project, removing stale results for confirmed deletions or exclusions and updating files that reappear during the scan. An unreadable folder is not treated as permission to delete its whole index.

## Why it matters

Routine restarts should reuse completed work, and search/trace results should reflect the current project after a branch switch or other offline changes.

## Limits

The faster restart is implemented for local-file and PostgreSQL indexes, not Qdrant. grepai still walks the project folders. This does **not** fix the separate background-startup timeout in #218 or guarantee instant startup. A deliberately restored file timestamp may require a forced content check.

The separate, older interrupted-write problem described in #311 is not claimed as fixed here. These checks verify that referenced data exists, not that all historical writes were transactional.

## How it was checked

- Reproduced the linked-directory regression with a copy of 230 real files whose hashes and timestamps still matched the saved index. Four files that had since changed were explicitly excluded from the unchanged test.
- Counted actual HTTP requests to a local fake embedding server. The old build made 3,516 requests on an unchanged restart; the corrected build made **zero on two consecutive restarts**, retaining all 230 files and 3,524 chunks. PostgreSQL workspace tests also made zero requests; a real content change still caused a request.
- Full PostgreSQL-backed race tests, existing startup/error-path tests, and local static checks passed before the correction was submitted for review. Native-platform CI is required as well. No fixed large-project speedup is promised.
- Counted PostgreSQL workspace tests also exercised a model-size rejection followed by a quiet restart, and repaired one file after its chunks were deliberately removed. No real GPU was used in these acceptance tests.


jeremyakers:
The unchanged linked-folder restart regression is fixed in 9f078403bbf4840cf0785db9dba5e27cbac167cb, after tests and fork-first Code Review in https://github.com/jeremyakers/grepai/pull/9. Before submission we replayed 230 unchanged real files: the old build made 3,516 embedding requests, while the corrected build made zero on two restarts with all 3,524 chunks and file ownership retained. Actual PostgreSQL workspace tests also made zero requests, and a changed-file control still caused an embedding request. Requests were counted by a loopback fake server; no GPU was used for these tests. This addresses the recovery bypass, not the separate general cache-plumbing issue.

jeremyakers:
Advanced to b7335d2cca2650b31738ac5531b15f6f15099ecd after current-head fork Code Review (#11), full PostgreSQL-backed tests, and counted CLI tests before publication. The fix now checks actual referenced chunks, recognizes legacy/Windows reference forms, preserves identifiers when the model splits a chunk, and retains zero embedding requests on unchanged linked-directory restarts. Real PostgreSQL workspace acceptance used 230 unchanged files / 3,524 chunks: requests 0/0 on repeated restarts, 0 for a timestamp-only change, +1 for a content change, and successful repair after missing chunks. The separately identified, pre-existing interrupted-write generation issue is tracked in #311 and is not claimed as fixed.

jeremyakers:
The reviewed correction is now installed in the combined fork build. On the real three-project workspace, the previous quiet starts took about 197 seconds and repeatedly processed 3,773 chunks. After catching up with real changes, a counted quiet start took 61.2 seconds with zero embedding attempts; a final normal, unproxied start took 61.8 seconds with zero new chunks. A before/after metadata comparison found no size, mtime, or indexed-membership changes across 36,093 tracked paths during those quiet checks. The counter also recorded one ancillary proxy error of unrecorded cause; it recorded no attempted, forwarded, or blocked embedding requests, and the final watcher runs without the proxy. This is observational startup timing, not a claim of a universal fixed speedup.

yoanbernabeu:
Heads-up: #300 and #306 are now on `main`, and this branch conflicts with them — `CHANGELOG.md`, `cli/watch.go`, `store/gob.go` and `trace/store.go`.

`store/gob.go` is the awkward one: #300 reworked the same file to gate persistence on a mutation counter and to hand out deep copies (`cloneChunk`/`cloneDocument`), which overlaps directly with the recovered-record validation here. Could you rebase on current `main`?

What landed, for reference:
- #300 — GOB stores persist only when dirty; per-project writer lock taken by `watchProjectWithEventObserver`, which passes the canonical root down.
- #306 — fail-closed watcher coverage; `watchProjectWithEventObserver` gained an `onFatal` parameter and delegates to `watchProjectWithEventObserverAndFence`.

One gotcha I hit rebasing #306, likely relevant here given the startup-scan focus: since #300, startup persistence writes nothing when the initial scan produced no content, so any test asserting that `index.gob` exists after scanning an empty fixture project now fails. Giving the fixture a real source file restores the intent.

Also worth a sanity check on your side once rebased: #308 and #300 both touch how per-file records are trusted on restart, and I would rather your zero-embedding-request measurement be re-run on the rebased branch than assume it still holds.

jeremyakers:
Updated this branch with current upstream main using a history-preserving merge. The GOB dirty-save/deep-copy changes and writer-lock/fatal-watcher behavior are retained; the extracted startup scan now also preserves the cancellation checks from main. Full PostgreSQL-backed race tests, vet and lint passed. I reran the measurement on this exact updated branch: two unchanged linked-directory restarts made zero embedding HTTP requests with both local-file storage and a PostgreSQL workspace, retaining all 230 fixture files and 3,524 chunks. Timestamp-only changes stayed at zero requests; changed content and missing chunks were handled correctly. MCP initialization/discovery also passed with mcp-go 1.0. Fork-first Code Review is complete in jeremyakers/grepai#13; the pre-existing interrupted-update issue noted there is tracked separately in #311, not claimed fixed by this branch refresh. New head: f41a567.

